PR #42, opened 2025-10-21
Wizard Smite Rework

imatsoup:
## About the PR

Changes Wizard smite to deal the exact amount of damage (in cellular) needed to put a target in crit using a new event instead of instagibbing the target.
## Why / Balance

Wizard has a lot of things that make it overbearing and feels-bad. This is one of those things. The spell should still be pretty strong, and dealing cellular does mean that the target will probably be out of commission for a bit, but at least they have a chance to come back now.
## Technical details

Wrote a new spell event to handle the smite spell. Changed the smite prototype to use the new spell event. Updated the description of Smite to reflect what it does now.
## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Wizard Smite now puts the target in critical condition rather than gibbing the target.

### DIFF
--- a/Content.Shared/Magic/SharedMagicSystem.cs
+++ b/Content.Shared/Magic/SharedMagicSystem.cs
@@ -4,8 +4,10 @@ using Content.Shared.Body.Systems;
 using Content.Shared.Charges.Components;
 using Content.Shared.Charges.Systems;
 using Content.Shared.Coordinates.Helpers;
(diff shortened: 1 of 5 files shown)


imatsoup:
Attached media didn't work </3

Atakku (approved):
Tested in game, works great, LGTM!

PR #81, opened 2025-11-02
Set the maxplayer limit of saltern-II to be somewhat similar to packed.

Freelanci:
## About the PR

Set the max player limit for saltern-II to 30, instead of 20 (which was a change I should have immediately rejected).

## Why / Balance

We're rarely below 20 players, meaning this map would rarely show up. This was meant to create more options until we hit the minimum limit for mid-pop maps. Instead, right now there's a pretty sizable window where the only map selections are amber, packed and elkridge, the exact thing I was trying to solve by adding in saltern.

## Requirements

**Changelog**

:cl:

- fix: Revert the requested change to Saltern-II's max player limit, as it is too low to bridge the gap to mid-pop maps.


Atakku (approved):
lgtm

arenaconspiracy:
I'm disappointed.

PR #112, opened 2025-11-10
Ticket magnitude change and fixes

SnappingOpossum:
## About the PR
Multiplies all ticket amounts by 10 and adjusts up some ore processor recipes to match the materials consumed. Removes ticket printing from unused recipes and non ore processor recipes.

## Why / Balance
Printing items like reinforced glass or especially plasteel should not be punished.
Tickets should not be sourced from the sheetifier.

## Technical details
Added ticket amount prototypes for 15, 20, 25, 30, 500, and 1000.
Added a ticketProtoId to every ticket printing recipe that lacked it, and modified the one used by diamonds.
Removed the printTicket and ticketProtoId from unused and non ore processor recipes.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: All ticket amounts have been multiplied by 10.
- tweak: Ore processor recipes that consume more materials now produce more tickets.
- fix: Certain non ore processor recipes no longer print tickets.

### DIFF
--- a/Resources/Prototypes/_Carpmosia/Entities/Objects/Specific/Salvage/ticket.yml
+++ b/Resources/Prototypes/_Carpmosia/Entities/Objects/Specific/Salvage/ticket.yml
@@ -38,6 +38,38 @@
 
 - type: entity
   parent: SalvageTicket
+  id: SalvageTicket15
+  suffix: 15
+  components:
+  - type: Stack
+    count: 15
+
+- type: entity
+  parent: SalvageTicket
+  id: SalvageTicket20
+  suffix: 20
+  components:
+  - type: Stack
+    count: 20
+
+- type: entity
+  parent: SalvageTicket
+  id: SalvageTicket25
+  suffix: 25
+  components:
+  - type: Stack
+    count: 25
+
+- type: entity
+  parent: SalvageTicket
+  id: SalvageTicket30
+  suffix: 30
+  components:
+  - type: Stack
+    count: 30
+
+- type: entity
+  parent: SalvageTicket
   id: SalvageTicket50
   suffix: 50
   components:
@@ -51,3 +83,19 @@
   components:
   - type: Stack
     count: 100
+
+- type: entity
+  parent: SalvageTicket
+  id: SalvageTicket500
+  suffix: 500
+  components:
+  - type: Stack
+    count: 500
+
+- type: entity
+  parent: SalvageTicket
+  id: SalvageTicket1000
+  suffix: 1000
+  components:
+  - type: Stack
+    count: 1000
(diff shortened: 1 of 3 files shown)


CarpBot (changes requested):
Awaiting approval from @Atakku, @Merrokitsune, @imatsoup

SnappingOpossum:
Since nothing uses just the printTicket with this I could also deprecate that, though I'm really not sure if I should.

imatsoup:
So, I was under the impression you were just going to look into increasing the amount of tickets something like reinforced glass/plasteel whatever gave you in relation to the materials used to print them, and I think I misunderstood the extent to which you wanted to do this. I don't think increasing the amount of tickets printed by a factor of 10 across the board is a good idea without also massively increasing the prices of everything in the ticket vendor.

imatsoup:
oh nevermind lol you did in fact do that

imatsoup:
I guess I still don't fully understand *why* we had to increase everything by a factor of 10 instead of just increasing the amount of tickets the aformentioned plasteel etc recipes gave?

imatsoup:
I should probably wait to review this when I'm not so tired!

SnappingOpossum:
> I guess I still don't fully understand _why_ we had to increase everything by a factor of 10 instead of just increasing the amount of tickets the aformentioned plasteel etc recipes gave?

Some of them would end up as fractional amount, mainly the reinforced glasses, since those take half a steel sheet each.

imatsoup:
> > I guess I still don't fully understand _why_ we had to increase everything by a factor of 10 instead of just increasing the amount of tickets the aformentioned plasteel etc recipes gave?
> 
> Some of them would end up as fractional amount, mainly the reinforced glasses, since those take half a steel sheet each.

see, I was able to work that out after thinking about it for a second, but hands type faster than my brain think tonight :wine_glass: . i shall review this properly tomorrow

imatsoup (approved):
Gave it some consideration, and this is probably fine. It does *look* like fairly ridiculous inflation at first glance, but things appeared to be scaled appropriately to my eyes.

Resources/Prototypes/_Carpmosia/Entities/Objects/Specific/Salvage/ticket.yml
```diff
@@ -51,3 +83,19 @@
...
+  - type: Stack
+    count: 500
+
+- type: entity
+  parent: SalvageTicket
+  id: SalvageTicket1000
+  suffix: 1000
+  components:
+  - type: Stack
+    count: 1000
```
Atakku, line 101:
```suggestion
    count: 1000

```

newline :3c

Merrokitsune (approved):
Looks good to me.

Atakku (changes requested):

Atakku:
nvm, odd, reviewing last bits now

Atakku (approved):
Gold gives the same amount of tickets as steel, but that can be addressed in a later balance PR once we have more idea on that.

CarpBot (approved):

PR #152, opened 2025-11-24
Move Captain's ID to vault

BevvRatBites:
Captain's spare id within the captain's room is a bit redundant and makes it so that when there is no captain, the heads of staff must break down the door to captain's room and destroy the locker to secure it.
This feels like ass, and when not done will lead to antags getting easy access to some of the most powerful equipment in the game as well as AA.
It being moved to vault still fulfills the function of having a spare id, having it easily secured, acting captains getting to retrieve their access roundstart and it being secure from maladaptive influences while avoiding before mentioned destruction of command infrastructure.

Ignore the five commits of me figuring out how to individually upload 2 prs at once I'm dumb.


CarpBot (changes requested):

Resources/Prototypes/Catalog/Fills/Lockers/heads.yml
```diff
@@ -28,7 +28,7 @@
   id: LockerFillCaptainNoLaser
   table: !type:AllSelector
     children:
-    - id: CaptainIDCard
+#    - id: CaptainIDCard - Moved to vault freezer so acting captains don't need to break down the door to captain's room to secure disk/alp/AA
```
imatsoup, line 31: For upstream file modifications, we usually have add specific comments with the feature name to help with merge conflicts and track things better. Here's a snippet from the dev-guide on discord. 
image

imatsoup:
I see the vision. I'm gonna approve workflows and think the idea over tomorrow when I'm more awake.

imatsoup:
Allllso you need to use the PR template. At the very least for changelogs, since we actually use that.
Template example from another PR:
image

arenaconspiracy (changes requested):
I see the vision. Please see an optometrist.

I am not stoked on letting the wayward RD, or QM, or whoever already powergames into the Deckard on green alert, powergaming into free all access round start.

Let's put on our thinking caps for a second, and think for a second.
Is the captain's spare locked behind captain's access because:

A) It gives you captain access.
B) It gives you all access.
C) It lets you become the captain.
D) It's incredibly powerful.
E) All of the above.
F) No reason :heart:

If there's no captain and you have to secure the disk, you grab the bridge fireaxe which is right there, then grab the mapped tertiary ID on the desk which is also right there, then open the locker.
This serves no advantage in the grand scheme of a shift. I am closing this pull request.

Feel free to contribute in the future.

PR #166, opened 2025-11-30
Add 'Teach x a lesson' objective

imatsoup:
## About the PR

Added 'Teach x a lesson' as an objective for Traitors to roll, an objective that requires a target be killed at least once.
## Why / Balance

Objectives that require traitors to engage in combat and murder are generally good for causing chaos and intrigue in the round. Round removal, however, is not as fun. Hopefully, having a murder objective that doesn't require round removal all the time means we get more chaos and violence (thus, more round intrigue and things for Sec to do) without increasing the amount of round removal too much (accidents can still happen, but Final Death is not mandatory).
## Technical details

Wrote a fairly simple objective condition and component. Added objective prototype in a new .yml file in _Carpmosia for traitor objectives we make, and added it to objectiveGroups.yml under the TraitorObjectiveGroupKill category. While other forks have implemented this objective (and probably done so better), this is not a port. I entirely took my cues from the existing KillPersonConditionSystem and Component.
## Media

https://github.com/user-attachments/assets/e0e4bb60-6ec4-48d1-806a-4fa55a5520fd

## Requirements

- [X] I have added media to this PR or it does not require an in-game showcase.

## Breaking changes

**Changelog**

:cl:
- add: Added a new Traitor objective,'Teach (crewmember) a lesson.' This objective requires the target be killed at least once during the shift.


CarpBot (changes requested):

Merrokitsune (approved):
Looks good to me.

arenaconspiracy (changes requested):
As far as I can tell this tracks _any death_ instead of deaths caused by the antagonist themselves, which is bad for many reasons that have been discussed in the past. You can see about reusing the kill credit system from deathmatch as mentioned in https://github.com/space-wizards/space-station-14/pull/40960 to narrow it down to damage directly caused by the traitor. It has it's limitations but is Good Enough for a first implementation.

imatsoup:
> As far as I can tell this tracks _any death_ instead of deaths caused by the antagonist themselves, which is bad for many reasons that have been discussed in the past. You can see about reusing the kill credit system from deathmatch as mentioned in [space-wizards/space-station-14#40960](https://github.com/space-wizards/space-station-14/pull/40960) to narrow it down to damage directly caused by the traitor. It has it's limitations but is Good Enough for a first implementation.

So there's a couple of issues I'm having in implementing this. Notably, there are circumstances where you *should* be given a greentext that aren't related to you completing your objective (IE, something deletes the target or the target cryos). The Kill objectives I'd been using as a guide use IsCharacterDeadIc, which works for the current implementation on this PR, but doesn't work for if we're trying to specify direct damage from the player being responsible. I'm not sure if there's an answer to this.

I also had trouble implementing a way to get it to mark the correct objective. The ninja pr just uses TryGetObjectiveComp to get the corresponding objective component, but that can only get the first objective, which means if we have multiple instances of Teach objectives, it won't properly update the right ones. I'm pretty confident there's a way around this, but haven't been able to find an implementation I like that doesn't have it's own tradeoffs. Making these objectives unique would still have the problem listed above.

Content.Server/_Carpmosia/Objectives/Systems/TeachPersonConditionSystem.cs
SnappingOpossum, line 1: I feel like this could benefit from `Entity<T>` references where applicable.

SnappingOpossum (changes requested):
If this objective can only be rolled once per antag, the issue with tracking who died can probably be left as a code comment for now. Otherwise, it should either be fixed or made to work like that.

imatsoup:
Closing due to stale and also not really being sure if this is the right way to go forward anyway

PR #235, opened 2026-01-15
No lowpop survival

Atakku:
## Requirements

- [ ] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [ ] I have added media to this PR or it does not require an in-game showcase.

## Breaking changes

**Changelog**


wavesneglected:
why john 25 pop

SnappingOpossum:
Concern added from issues raised in discord:
1: This is higher than nuke ops, which we often already don't have enough ready for.
2: This leaves the only lowpop gamemodes as traitor and deathmatch (not in secret pool), with extended as a fallback if, or rather when, that fails.
Focus could instead be put on improving survival or adding new lowpop gamemodes to fill the gap

PR #239, opened 2026-01-17
Improve AI navmap perf

ArtisticRoomba:
## About the PR
Improves navmap perf.

Idle alloc rate from 500+ KiB/frame to ~40 KiB/frame, also improves moving drawing perf slightly but there are still some issues that are hard to resolve due to the nature of the implementation.

## Why / Balance
This is something I did in https://github.com/space-wizards/space-station-14/pull/41889/changes/ca016ab6962bc4a849d2f5bdb3b0d4f258dcc6e9 as like a baseline poke into cleaning up the PR. In general I don't think this is right right approach as in order to make this usable for AI in the future you would basically have to create your own bespoke AI-navmap implementation that properly offsets things. I'd have to think about it more.  But that is your choice to make.

Just bringing this down here because nobody seems to have picked it up and it's free.

Code is MIT.

## Technical details
Creating a ValueList every single draw call is not performant, the backing implementation is still an array. Just re-use the array, you don't need to create one every time you draw.

## Media
image

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an in-game showcase.

## Breaking changes
n/a

**Changelog**
:cl:
- fix: Improved performance for the AI navmap when idle on lower end machines.

### DIFF
--- a/Content.Client/Silicons/StationAi/StationAiOverlay.cs
+++ b/Content.Client/Silicons/StationAi/StationAiOverlay.cs
@@ -32,9 +32,17 @@ public sealed class StationAiOverlay : Overlay
     private readonly NavMapControl _navMap = new(); // Carpmosia-edit - AI Navmap
 
     private readonly OverlayResourceCache<CachedResources> _resources = new();
-    private Dictionary<Color, Color> _sRGBLookUp = new(); // Carpmosia-edit - AI Navmap
 
-    private float _updateRate = 1f / 30f;
+    // Carpmosia-start - AI Navmap
+    private readonly Dictionary<Color, Color> _sRgbLookUp = new();
+    private static readonly RenderTargetFormatParameters RenderTargetFormatParameters = new(RenderTargetColorFormat.Rgba8Srgb);
+
+    private static readonly List<Vector2> TileLinesToDraw = [];
+    private static readonly List<Vector2> TileRectsToDraw = [];
+
+    private const float UpdateRate = 1f / 30f;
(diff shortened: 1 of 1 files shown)


CarpBot (changes requested):

imatsoup (approved):
very much appreciated

Content.Client/Silicons/StationAi/StationAiOverlay.cs
```diff
@@ -81,7 +92,9 @@ protected override void Draw(in OverlayDrawArgs args)
             _navMap.AiFrameUpdate((float) _timing.FrameTime.TotalSeconds, gridUid); // Carpmosia-edit - AI Navmap
             if (_accumulator <= 0f)
             {
-                _accumulator = MathF.Max(0f, _accumulator + _updateRate);
+                // Carpmosia-start - AI Navmap
+                _accumulator = MathF.Max(0f, _accumulator + UpdateRate);
+                // Carpmosia-end - AI Navmap
```
Atakku:
```suggestion
                _accumulator = MathF.Max(0f, _accumulator + UpdateRate); // Carpmosia-edit - AI Navmap
```

Atakku (approved):
lgtm, just could use an edit annotation here instead

CarpBot (approved):

PR #242, opened 2026-01-21
Give a hand to supplybot

korczoczek:
## General information

Adds hands to supplybot, no combat mode, and only papers.

## Why / Balance

Finally lets them look inside and interact with their own inventory, which is important given their purpose as a delivery robot. Hands are limited to Papers and Deliverables only via a whitelist.

## Technical details

Adds Hands with whitelist, adds ComplexInteraction

## Media

https://github.com/user-attachments/assets/7ff96fcb-84f4-4629-a609-30f5813b43c7

## Changelog

:cl:
- tweak: Gave supplybot hands for checking its own inventory and picking up papers, letters and packages


CarpBot (changes requested):

imatsoup:
Think this might require some more discussion on what we want supplybots to be. I'm not opposed to giving them something, given that the role isn't really something I see people take, but I don't know about letting them pick up whatever. Maybe whitelisting the hand for mail or something would be more appropriate.

korczoczek:
> Think this might require some more discussion on what we want supplybots to be. I'm not opposed to giving them something, given that the role isn't really something I see people take, but I don't know about letting them pick up whatever. Maybe whitelisting the hand for mail or something would be more appropriate.

This may actually be better than the idea i had, which was figuring out how to set the blacklist for everything, making the hand only functional for opening it's own inventory.

The paper/mail hand idea is good

Atakku:
> > Think this might require some more discussion on what we want supplybots to be. I'm not opposed to giving them something, given that the role isn't really something I see people take, but I don't know about letting them pick up whatever. Maybe whitelisting the hand for mail or something would be more appropriate.
> 
> This may actually be better than the idea i had, which was figuring out how to set the blacklist for everything, making the hand only functional for opening it's own inventory.
> 
> The paper/mail hand idea is good

there is an opposite of blacklist, called whitelist :godo:

korczoczek:
> there is an opposite of blacklist, called whitelist :godo:

Yea, but i'm 99% sure whitelists are ignored if set to empty :derp:

korczoczek:
added the paper-only whitelist

korczoczek:
turns out it is indeed complexinteraction that lets you pick stuff off the ground 👍

korczoczek:
This does mean that they can now interact with vending machines and consoles, but they also have immutable Cargo-only access. So that shouldn't be a problem

imatsoup (approved):
works

Atakku (changes requested):
- paper visual is stuck on right hand, but idk how to fix that
- you can destroy yourself by touching the light bulb enough times, dont borgs have immunity to that usually?
- should probably have cargo access to access the mailbox
- it can already interact with food vends and etc, would be funny if it could also bring those foods, at least most of them to other people

korczoczek:
> * paper visual is stuck on right hand, but idk how to fix that
> * you can destroy yourself by touching the light bulb enough times, dont borgs have immunity to that usually?
> * should probably have cargo access to access the mailbox
> * it can already interact with food vends and etc, would be funny if it could also bring those foods, at least most of them to other people

- not my problem, it's an issue with hands code, or even an engine problem
- ok, i'll fix that
- they already have cargo access
- they already can drag stuff around

korczoczek:
@Atakku solved the burned hands issue, and also here's a showcase of a supplybot doing everything else that was asked of it

https://github.com/user-attachments/assets/a504eca4-9718-4a6e-bd2e-80420e4cf0c9

Atakku (approved):
mrow :3

CarpBot (approved):

PR #248, opened 2026-01-27
Shared Blood Bound Objectives

imatsoup:
## General information

Changes Blood Bound rule system to add the objectives of the original blood bound to the converted blood bound's mind. Changes Steal Condition system to allow us to count either blood bound holding onto a steal objective as a green-text for them both.
## Why / Balance

A requested quality of life improvement for the antagonist. 
## Technical details

Changes StealConditionSystem, adding a new function to check for the presence of BloodBoundComponent, and then return a list of all 'allied' entities to iterate through when checking steal conditions. Slightly changed BloodBoundRuleSystem to add all of the original bloodbound's objectives to the convert.
## Changelog

:cl:
- tweak: Converted Blood Bound will now have the objectives of the original Blood Bound.
- tweak: For Blood Bound steal objectives, either Blood Bound carrying the objective item will count as a greentext for them both.

### DIFF
--- a/Content.Server/Objectives/Systems/StealConditionSystem.cs
+++ b/Content.Server/Objectives/Systems/StealConditionSystem.cs
@@ -105,60 +105,60 @@ public sealed class StealConditionSystem : EntitySystem
         // Carpmosia-start - Shared Blood Bound Objectives
         foreach(var ally in CheckAllies(mind))
         {
-            if (!_containerQuery.TryGetComponent(ally, out var currentManager))
-                return 0;
+        if (!_containerQuery.TryGetComponent(ally, out var currentManager))
+            return 0;
+        // Carpmosia-end - Shared Blood Bound Objectives
+        _countedItems.Clear();
 
-            _countedItems.Clear();
-
-            //check stealAreas
-            if (condition.CheckStealAreas)
+        //check stealAreas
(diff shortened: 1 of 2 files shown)


CarpBot (changes requested):

SnappingOpossum:
I feel like something more generic would be a better solution here rather than just trying to connect all possible components that could have you allied.

Atakku:
what happens to objectives when either of the bloodbound gets mindshielded?

imatsoup:
> I feel like something more generic would be a better solution here rather than just trying to connect all possible components that could have you allied.

Been struggling a bit to figure out something more generic/scalable. Any suggestions? The least ugly solution I could think of on my own is having a component track allies separately and then look at that in the steal condition system, but that doesn't feel like a very good solution.

imatsoup:
> what happens to objectives when either of the bloodbound gets mindshielded?

I hadn't thought to check that initially and the added objectives aren't removed properly. Going to make a change to the underlying system to track which objectives are added and remove all of those objectives properly.

Atakku:
i guess with mindshielding removed from blood bounds this is no longer an issue xD

Atakku (approved):
mrrp miau

CarpBot (approved):

PR #254, opened 2026-02-02
Added a Mail Network to [Carp]Saltern

korczoczek:
## General information

Added a Mail Network to :carp: Saltern :carp:

## Why / Balance

Brings the map up to modern standards, it was one of the only maps in-rotation missing a mailing system. Even Feint has one (more or less).

## Technical details

1000 years sentence in mapping mines

Tested it by emagging mailing units and mailing myself randomly, didn't get stuck or flung out or mailed to the wrong place so i think we're gucci

## Media

Bar
obraz

Bridge
obraz

Cargo
obraz

Court
obraz

Engineering
obraz

Library
obraz

Medical
obraz

Observatory
obraz

Science
obraz

Security
obraz

## Changelog

:cl:
- add: On [Carp]Saltern: Added a Mailing System connected to various points around the station.

### DIFF
--- a/Resources/Maps/_Carpmosia/saltern.yml
+++ b/Resources/Maps/_Carpmosia/saltern.yml
@@ -1,11 +1,11 @@
 meta:
   format: 7
   category: Map
-  engineVersion: 268.0.0
+  engineVersion: 272.0.0
   forkId: ""
   forkVersion: ""
-  time: 12/29/2025 14:05:09
-  entityCount: 17233
+  time: 02/02/2026 19:46:25
+  entityCount: 17495
 maps:
 - 1
 grids:
@@ -7095,6 +7095,8 @@ entities:
     - type: NavMap
     - type: ImplicitRoof
     - type: ExplosionAirtightGrid
+    - type: TileHistory
+      chunkHistory: {}
   - uid: 4
     mapInit: true
     paused: true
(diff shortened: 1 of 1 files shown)


CarpBot (changes requested):
Awaiting approval from @Atakku, @wavesneglected, @imatsoup

imatsoup (approved):
i'm into it, conceptually

korczoczek:
Also, test fails are completely unrelated, github actions decided to have an outage right as i was making the pr

Cogrem:
I think the one at medical is a little cumbersome, as it's right in the middle.
If I can suggest anything, it would be to shorten the table in the top left, move the stuff that's on it down, and place the mailing unit here.


Again, I don't have my word to say here so you can ignore that.

korczoczek:
> I think the one at medical is a little cumbersome, as it's right in the middle. If I can suggest anything, it would be to shorten the table in the top left, move the stuff that's on it down, and place the mailing unit here.
> 
> Again, I don't have my word to say here so you can ignore that.

That mailing unit is already replacing a table segment, that is the plasmaglass table previously extended to where the unit now stands

wavesneglected (approved):
Did the same test, looks good to me.

Atakku (approved):
mrow

CarpBot (approved):

PR #267, opened 2026-02-26
Modify secret weights for Carpmosia

SnappingOpossum:
## General information

Makes traitor more common, nuke ops less common.

## Why / Balance

Requested by Celeste.

From Celeste: Nukeops tends to be a greenshift until it isnt, trying to up traitor rates also makes sure low pop has a gamemode thats player driven, instead of event driven, this also equalized the gamemodes to be similar in weights, instead of nukeops being the most common out of unique modes

## Technical details

Add a weightedRandom, make secret_weight_prototype use it.

## Changelog

:cl:
- tweak: For the secret rotation, traitor has been made more common and nuke ops has been made significantly less common.


SnappingOpossum:
Forgot the title, whoops

CarpBot (changes requested):

Resources/Prototypes/_Carpmosia/secret_weights.yml
```diff
@@ -0,0 +1,8 @@
+- type: weightedRandom
+  id: CarpmosiaSecret
+  weights:
+    Traitor: 70
+    Nukeops: 10
+    Survival: 10
+    Zombie: 5
+    Revolutionary: 5
```
arenaconspiracy, line 8: This is unprecedented. This is mental. What was celestes reasoning to make nukeops vanishingly rare? Why bury zombies and revolutionaries even further? What's the point??? Carpmos is already a greenshift hell dimension and this is just making it worse, i only log on to play security these days and this might genuinely get me to quit for good. I would like to hear clear reasoning behind these weights, otherwise it just feels like it's making the game monotone for no good reason.

arenaconspiracy (changes requested):
The changelog entry needs to say specifically the new relations of the gamemode weights, just saying they're different now is vague and unhelpful. It's the _secret_ rotation.

Resources/Prototypes/_Carpmosia/secret_weights.yml
SnappingOpossum: Zombies, revs, and survival aren't made that much rarer by this, came chance as before wizard was disabled.

imatsoup:
Do you think you could get Celeste's reasoning and add it to the Why/Balance portion of the PR for us?

SnappingOpossum:
> Do you think you could get Celeste's reasoning and add it to the Why/Balance portion of the PR for us?

I don't have exact words but it was about making nukeops rarer so that it's more of a mix up when it does happen. I'll toss a discussion label on this though, since all the values here are subject to change.

Atakku:
weights before wizard disabling
image

weights after wizard disabling
image

weights in this pr
image

Atakku:
ok this hasn't really went anywhere much, and I think we should return to this later in a new PR, after we see how https://github.com/carpmosia/carpmosia/pull/275 works out

PR #270, opened 2026-03-02
Make dead/crit bodies much harder to pull

ultrafiche:
## General information
Applies a higher friction modifier to prone bodies that are incapacitated (crit/dead), making them much harder to move.

## Why / Balance
Gives more of a job to our poor paramedics, who will now have a greater incentive to stabilize (or even treat for the especially proactive) and transport patients in the field. Also encourages more strategic methods of dealing with casualties, ad hoc or otherwise; an example I gave on discord was a bystander using a chair to move a body.

Should lead to overall more interesting RP vs dying -> being dragged to med in like 30 seconds -> being revived -> repeat

## Technical details
This is my first time touching C# code so this might be far from best practice or conventions :godo:
- Adds the `Incapacitated` and `LimpFrictionMod` datafields to `StandingStateComponent`.
- Adds the actual logic to `StandingStateSystem` to apply the correct friction modifier, based on `Incapacitated` which is set according to a `MobStateChangedEvent`.
    - I would prefer to handle this in something like `MobStateSystem` but StandingState overrides any friction modifiers when prone so this just adds another path for that override to take.
## Media
https://github.com/user-attachments/assets/d51d9900-a44f-4ce8-8b63-7a609eb0577b

## Breaking changes
Only public changes are the new datafields in `StandingStateComponent`.

## Changelog

:cl: microfiche
- add: Moving incapacitated bodies is now much more cumbersome.

### DIFF
--- a/Content.Shared/Standing/StandingStateComponent.cs
+++ b/Content.Shared/Standing/StandingStateComponent.cs
@@ -14,11 +14,20 @@ namespace Content.Shared.Standing
         [DataField, AutoNetworkedField]
         public bool Standing { get; set; } = true;
 
+        [DataField, AutoNetworkedField]
+        public bool Incapacitated { get; set; } = false;
+
         /// <summary>
         /// Friction modifier applied to an entity in the downed state.
         /// </summary>
         [DataField, AutoNetworkedField]
         public float DownFrictionMod = 0.4f;
+
+        /// <summary>
+        /// Friction modifier applied to an entity in an incapacitated (crit/dead) state.
+        /// </summary>
+        [DataField, AutoNetworkedField]
+        public float LimpFrictionMod = 3f;
 
         /// <summary>
         ///     List of fixtures that had their collision mask changed when the entity was downed.
--- a/Content.Shared/Standing/StandingStateComponent.cs
+++ b/Content.Shared/Standing/StandingStateComponent.cs
@@ -14,8 +14,10 @@ namespace Content.Shared.Standing
         [DataField, AutoNetworkedField]
         public bool Standing { get; set; } = true;
 
+        // Carpmosia-start - make dead/crit bodies much harder to pull
         [DataField, AutoNetworkedField]
         public bool Incapacitated { get; set; } = false;
+        // Carpmosia-end - make dead/crit bodies much harder to pull
 
         /// <summary>
         /// Friction modifier applied to an entity in the downed state.
@@ -23,11 +25,13 @@ namespace Content.Shared.Standing
         [DataField, AutoNetworkedField]
         public float DownFrictionMod = 0.4f;
 
+        // Carpmosia-start - make dead/crit bodies much harder to pull
         /// <summary>
         /// Friction modifier applied to an entity in an incapacitated (crit/dead) state.
         /// </summary>
         [DataField, AutoNetworkedField]
         public float LimpFrictionMod = 3f;
+        // Carpmosia-end - make dead/crit bodies much harder to pull
 
         /// <summary>
         ///     List of fixtures that had their collision mask changed when the entity was downed.
--- a/Content.Shared/Standing/StandingStateComponent.cs
+++ b/Content.Shared/Standing/StandingStateComponent.cs
@@ -14,11 +14,6 @@ namespace Content.Shared.Standing
         [DataField, AutoNetworkedField]
         public bool Standing { get; set; } = true;
 
-        // Carpmosia-start - make dead/crit bodies much harder to pull
-        [DataField, AutoNetworkedField]
-        public bool Incapacitated { get; set; } = false;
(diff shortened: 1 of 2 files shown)


CarpBot (changes requested):

Content.Shared/Standing/StandingStateSystem.cs
```diff
@@ -70,6 +86,16 @@ private void OnEndClimb(Entity<StandingStateComponent> entity, ref EndClimbEvent
...
     }
 
+    private void OnMobStateChanged(Entity<StandingStateComponent> entity, ref MobStateChangedEvent args)
+    {
+        if (args.NewMobState == MobState.Critical || args.NewMobState == MobState.Dead) {
+            entity.Comp.Incapacitated = true;
+            return;
+        }
+
+        entity.Comp.Incapacitated = false;
```
SnappingOpossum: Could just set it to the outcome of `args.NewMobState == MobState.Critical || args.NewMobState == MobState.Dead`

SnappingOpossum (dismissed):
Needs modification comments.

ultrafiche:
FUCKKK I FORGOT THE COMMENTS

Content.Shared/Standing/StandingStateSystem.cs
```diff
@@ -51,14 +54,31 @@ private void OnRefreshFrictionModifiers(Entity<StandingStateComponent> entity, r
...
+        if (entity.Comp.Standing)
+            return;
+
+        if (entity.Comp.Incapacitated) {
+            args.Modifier *= entity.Comp.LimpFrictionMod;
+            return;
+        }
+        // Carpmosia-end - make dead/crit bodies much harder to pull
+
+        args.Modifier *= entity.Comp.DownFrictionMod;
```
korczoczek, line 80: move the comment so it's under all your changes

Content.Shared/Standing/StandingStateComponent.cs
```diff
@@ -14,12 +14,25 @@ public sealed partial class StandingStateComponent : Component
...
         /// </summary>
         [DataField, AutoNetworkedField]
         public float DownFrictionMod = 0.4f;
 
+        // Carpmosia-start - make dead/crit bodies much harder to pull
+        /// <summary>
+        /// Friction modifier applied to an entity in an incapacitated (crit/dead) state.
+        /// </summary>
+        [DataField, AutoNetworkedField]
+        public float LimpFrictionMod = 3f;
```
korczoczek, line 31: nitpick, but i'd move the new variables together, so they're encompassed by a single carp comment pair

korczoczek (changes requested):
Code works as advertised, just have some minor nitpicks about the code

Content.Shared/Standing/StandingStateComponent.cs
ultrafiche: makes sense, wasnt sure if datafields of the same type should be kept near each other

korczoczek (approved):

imatsoup:
Feel unsure about it. A live body prone being significantly faster to drag than a dead one feels weird diegetically, traitors now have a much harder time dragging dead bodies to hide them. I suppose it helps discourage thoughtlessly running into danger as without a paramedic, you are more likely to rot before you reach medical? Not sure how this will feel with the changes to med that are planned in the future from upstream either.

imatsoup (approved):
Decided to seek some counsel and lets give it a shot. If it turns out to be overbearing or break Offmed months later or something, we have the capacity to revert it down the line.

ultrafiche:
Diegetically, a body without muscle tone is very unwieldy (irl first responders are trained in emergent and urgent maneuvers specifically because limp bodies are hard to move, otherwise a stretcher is used ASAP). I also just didn’t want to make arresting someone suckkkk. 

Regarding traitors, I personally feel that if a traitor has killed someone and they need to move the body NOW or risk getting caught, their plan has seriously gone wrong and they should face some friction for having to do something so brazen. They could snatch a chair or something beforehand if need be.

Also I am absolutely not in the loop regarding offmed lol if this does interfere with it's balance feel free to rip it out.

arenaconspiracy:
I think it's worth exploring handcuffs lowering your tile friction to make dragging prisoners not miserable and that entirely separately from this skyrockets security QoL. Something like security restraints having a better friction modifier than makeshift restraints would be good.
It's easier to drag cuffed people when they're crawling than standing, and I've had to ask someone "Er, can you lay down for a second." so I could faster get them to the emergency shuttle after it docked and that's kind of fucked. All of that is tangential and out of scope for the PR though.

Atakku (commented):
looks good, works good, is good

Atakku (approved):
miau

CarpBot (approved):

PR #272, opened 2026-03-05
Add support and make Common radio read-only on headsets

arenaconspiracy:
A low fidelity port of https://github.com/impstation/imp-station-14/pull/2931
I consider this barebones implementation [canonical (definition 3)](https://en.wiktionary.org/wiki/canonical) therefore indefensible under license disputes :trollface:.
## General information
Adds support for radio channel prototypes being intercom-only, essentially being read-only to headsets.
Marks Common radio as being intercom-only.

## Why / Balance
Communication methods including departmental radios, console announcements, holopads (especially emergency broadcasts), faxes, et cetera; are neglected and outclassed by the availability and ubiquity of Common radio, delivering information to the most people with the most convenience.

The availability of Common radio lends itself to chat spam with LRP hogwash at all times of day, and specific information gets blasted to 20-40 people to whom it is irrelevant (e.g. the whole station hearing that the chemvend needs a restock).

Limiting Common radio encourages people to seek out intercoms in known locations, and emphasizes/enables the Station AI,  Head of Personnel, and the Captain relaying information between departments.

## Technical details
Adds a bool datafield to RadioChannelPrototype for intercom-only behavior, defaulting to false. Sets it to true on Common.
Adds a check to RadioSystem doing an AND check on intercom-only and presence of HeadsetComponent on the radio message source, cancels the radio message event if both are true.

## Media
https://github.com/user-attachments/assets/b5a57a25-dd6a-46c2-939f-6b5395bc10c8

## Breaking changes
N/A

## Changelog
:cl:
- tweak: Headsets can no longer speak on Common radio, you need an intercom or other means to use it. Other radio channels are unaffected.


CarpBot (changes requested):

arenaconspiracy:
Waiting for the hate this will get with open arms.

Resources/Prototypes/radio_channels.yml
```diff
@@ -112,4 +113,4 @@
   frequency: 2003
   color: "#ff2222"
   # long range since I don't wanna make a special xenoborg telecomm server
-  longRange: true
\ No newline at end of file
+  longRange: true
```
imatsoup: believe it needs the same carpmosia-edit comment as above

arenaconspiracy: No it doesn't, the diff is from an EOL getting messed with by writing the file in vim.

arenaconspiracy: No it doesn't, the diff is from an EOL getting messed with by writing the file in vim.

arenaconspiracy:
Technically the boolean could/should be called something else since cyborg chassis/AI cores/anything else with radio capabilities and no HeadsetComponent can still speak on the channel, making the name IntercomOnly a misnomer, but I don't really care.

Content.Shared/Radio/RadioChannelPrototype.cs
```diff
@@ -35,4 +35,12 @@ public sealed partial class RadioChannelPrototype : IPrototype
     /// </summary>
     [DataField("longRange"), ViewVariables]
     public bool LongRange = false;
+
+    // Carpmosia-start - intercom only channels
+    /// <summary>
+    /// Is this channel read-only to headsets
+    /// </summary>
+    [DataField("intercomOnly"), ViewVariables]
```
SnappingOpossum: Matches autogenerated name.
```suggestion
    [DataField, ViewVariables]
```

Content.Server/Radio/EntitySystems/RadioSystem.cs
```diff
@@ -145,6 +145,11 @@ public void SendRadioMessage(EntityUid messageSource, string message, RadioChann
             if (attemptEv.Cancelled)
                 continue;
 
+           // Carpmosia-start - intercom only channels
```
SnappingOpossum:
```suggestion
       // Carpmosia-start - intercom only channels
```

Content.Shared/Radio/RadioChannelPrototype.cs
```diff
@@ -35,4 +35,12 @@ public sealed partial class RadioChannelPrototype : IPrototype
...
+    public bool IntercomOnly = false;
```
imatsoup: Yeah this should probably be called something like IsHeadsetOnly or something. I could update the variable name to something if you want

SnappingOpossum (dismissed):
I said it elsewhere so I'll say it here as well, I feel this would be better handled by a whitelist/blacklist type system, though this implementation does accomplish what it tries to achieve.

Content.Shared/Radio/RadioChannelPrototype.cs
SnappingOpossum: Technically just allows it on headsets when `false` as is, so `IsHeadsetOnly` wouldn't fit well.

arenaconspiracy: NoHeadsets :julia:

SnappingOpossum: Could be `HeadsetReadOnly` or something like that.

Content.Shared/Radio/RadioChannelPrototype.cs
arenaconspiracy: I'm completely aware it matches the auto-gen but I kept it in for the sake of matching the rest of the file, I was on the fence on doing it one way or the other so this gives me an excuse to go with my better judgement.

Content.Shared/Radio/RadioChannelPrototype.cs
arenaconspiracy: `HeadsetReadOnly` would describe exactly what it does and leaves room for `[Component]ReadOnly` in the future (albeit again that would better be done wirh whitelists)

arenaconspiracy:
I do think it would be cleaner done with a blacklist but in my opinion it's completely unnecessary when the use-case of blacklisting anything but headsets isn't actually desired.

SnappingOpossum:
> I do think it would be cleaner done with a blacklist but in my opinion it's completely unnecessary when the use-case of blacklisting anything but headsets isn't actually desired.

I mean currently xenoborgs can't speak on the mothership channel because of how their ActiveRadio and IntrinsicTransmitters are defined, though in that sense you could instead apply this to how the encryption keys themselves are defined.

Atakku:
The chsngelog is confusing, it sounds like headsets CANT use other radios anymore. Perhaps reword it to Headsets can't talk on common anymore

arenaconspiracy:
Is that better

Atakku:
> > I do think it would be cleaner done with a blacklist but in my opinion it's completely unnecessary when the use-case of blacklisting anything but headsets isn't actually desired.
> 
> I mean currently xenoborgs can't speak on the mothership channel because of how their ActiveRadio and IntrinsicTransmitters are defined, though in that sense you could instead apply this to how the encryption keys themselves are defined.

does this PR specifically affect that or

arenaconspiracy:
tmp

Doesn't.

imatsoup (approved):
I think this is ultimately a good change for a lot of different reasons. Let's do it.

Atakku:
i suppose this isn't the best approach but it is for now because i cba to implement anything better

Atakku (approved):
mrrp miau

CarpBot (approved):

PR #282, opened 2026-03-12
Remove current blood bound mindshield deconversion

imatsoup:
## General information

Blood bound converts are no longer deconverted on mindshield.
## Why / Balance

Currently blood bound are one of the most vulnerable antagonists. Their entire gimmick relies on trading all the equipment and abilities of other antagonists for one guaranteed ally that can be taken away through the application of an implant that security has no reason to withhold under most circumstances. Mental shielding is something that security *has* to be able to apply to people for any reason due to the existence of revolutionaries, and rather than trying to carve out some sort of exception in metagaming rules for blood bound that adds a ton more work for admins, I'd rather address the problem in this way. 

This does mean that, as security, there is currently no counterplay if blood bound decide to act overtly, which is a loss. I am currently exploring ways to deconvert the blood bound that are less easy/more costly or possibly giving blood bound other advantages that security can take away through mental shielding so that they have an incentive to maintain some stealth.
## Technical details

Remove OnBloodBoundMindShielded and relevant deconversion string, as well as relevant system call in SharedRevolutionarySystem. 
## Media

image

## Changelog

:cl:
- tweak: Blood bound converts will no longer be deconverted when implanted with a mindshield.

### DIFF
--- a/Content.Shared/Revolutionary/SharedRevolutionarySystem.cs
+++ b/Content.Shared/Revolutionary/SharedRevolutionarySystem.cs
@@ -12,7 +12,6 @@ namespace Content.Shared.Revolutionary;
 
 public abstract class SharedRevolutionarySystem : EntitySystem
 {
-    [Dependency] private readonly SharedBloodBoundSystem _bloodBoundSystem = default!; // Carpmosia-edit - Harmony Blood Bound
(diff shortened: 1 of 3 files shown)


CarpBot (changes requested):

Atakku (approved):
mrrp miau

CarpBot (approved):

PR #289, opened 2026-03-17
Patch out old vent creatures

imatsoup:
## General information

Patches out changes made for the old version of vent creatures and fully reverts the relevant lines to upstream.
## Why / Balance

Seems like a better version of the same idea. If we end up liking our interpretation more, things can be reverted easily enough.
## Technical details

Compared our C# files to theirs and chose theirs.
## Changelog

:cl:
- tweak: Carp version of the vent critters event replaced by the new upstream vent hordes event

### DIFF
--- a/.github/workflows/build-test-debug.yml
+++ b/.github/workflows/build-test-debug.yml
@@ -51,7 +51,7 @@ jobs:
       shell: pwsh
       run: |
         $env:DOTNET_gcServer=1
-        dotnet test --no-build --configuration DebugOpt Content.IntegrationTests/Content.IntegrationTests.csproj -- NUnit.ConsoleOut=0 NUnit.MapWarningTo=Failed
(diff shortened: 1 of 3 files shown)


CarpBot (changes requested):

Atakku:
With the horde having a visible indication of something about happening, and the hilarious popout visual of critters flying out, I think the upstream horde implementation is good as-is and my modifications can be fully removed (I was considering that maybe we should keep the area spread, but that would kinda ruin the hilarious explosion visual, I initially implemented it because 6 spiders spawning all in one tile was meh and there was no way a person could handle that attack (but in here they have time to react and run away from when the  thing starts shaking))

Atakku (approved):
mrrp miau

CarpBot (approved):

PR #290, opened 2026-03-17
Rename common channel and redo common key name and desc

### LINKED ISSUE #285: Common encryption key readability issues
As pointed out by Void in the [discord](https://discordapp.com/channels/1404602275401568347/1479692921690394797/1482875858934042635), the common encryption key is now out of line with the other encryption keys. I'd like to see the following changes they suggested to the common comms key implemented. I'll likely get to them eventually but making it an issue like this helps keep it organized and present

- the "Common" radio channel should be renamed to the "Intercom" channel, since it requires an intercom to talk on
- the encryption keys for headsets should have their names and descriptions edited to be Intercom Receiver Keys or something similar, so that they are distinct from regular encryption keys
- (from arenaconspiracy) changing the inspect description on headsets to replace "; for Common" with "read only: Common", which helps make the change clearer
- since the encryption keys would be changed to receiver keys, we could make access to the Intercom/Common channel inherent to intercoms on construction, which would make the behavior of the common/receiver keys more consistent across headsets and intercoms.

They made some other suggestions related to handheld radios, but I'd like to evaluate those separately.

imatsoup:
## General information

Changes the description generated by read-only radio channels to indicate that they are read-only rather than displaying the keycode associated with the channel, and updates the name and description of the common key to be more accurate to what it actually does now.
## Why / Balance

The common channel now has unique behavior (being a read-only channel) and has no indication that it would behave that way, either on the encryption keys or the interface. This fixes that and partially resolves https://github.com/carpmosia/carpmosia/issues/285.
## Technical details

Changes to yaml and ftl strings.
## Media

 Updated radio examine description and name
image

## Changelog

:cl:
- tweak: Common radio channel has been renamed to Intercom. 
- tweak: Radio channel descriptions now differentiate between channels that are read-only and normal radio-channels.


CarpBot (dismissed):

arenaconspiracy (changes requested):
This is not the right way to do this. The loc string for the channel name should just be `Intercom`, and the _examine text_ should state that it's read only. Ergo, instead of `{$key} for {$name}` it should be `Read only: {$name}`. The current implementation makes the label over twice as long as any other radio channel, which ends up taking up valuable real estate in the chat box. The examine system should be able to check for the same NoHeadset bool on the channel prototype trivially.

arenaconspiracy:
Also add media dawg.

Resources/Prototypes/Entities/Objects/Devices/encryption_keys.yml
```diff
@@ -14,8 +14,8 @@
 - type: entity
   parent: EncryptionKey
   id: EncryptionKeyCommon
-  name: common encryption key
-  description: An encryption key used by anyone.
+  name: intercom receiver key
+  description: An encryption key used by anyone. Can only receive messages from the public intercom channel.
```
SnappingOpossum: Lacks comments, still is an encryption key that can send outgoing messages since intercoms use this for the channel.

Resources/Locale/en-US/headset/headset-component.ftl
```diff
@@ -4,7 +4,9 @@ chat-radio-message-wrap-bold = [color={$color}]{$channel} [bold]{$name}[/bold] {
 
 examine-headset-default-channel = Use {$prefix} for the default channel ([color={$color}]{$channel}[/color]).
 
-chat-radio-common = Common
+# Carpmosia-start - Common to Intercom
+chat-radio-common = read-only: Intercom
+# Carpmosia-end - Common to Intercom
```
SnappingOpossum: Single # comments in ftl references the following line.
```suggestion
# Carpmosia-edit - Common to Intercom
chat-radio-common = read-only: Intercom
```

SnappingOpossum (changes requested):

wavesneglected:
Frankly im unsure if we should move to make this change and continue work on this now to the future. popular opinion is against the common channels and as much as id love to keep it like this it might be better to plan for a revert, address the issues that are present in the system and give alternative methods of communication, then plan a more worked out re-implementation.

arenaconspiracy:
There isn't really a point in adding More Alternative Communication while R/W Common is present I think, gauging the utility and necessity of new methods is impossible when Ubiquitous Common blows it out the water, plus I think it's missing the point that RO Common doesn't need a replacement when _the whole point_ is adding friction to communications.
The current facets of RO Common can be augmented piecemeal and a single PR can be made reverting every pertinent change at once, so it's pretty arbitrary how many separate PRs are made for it when reverting them all isn't made any harder.

arenaconspiracy:
Piecemeal progress notwithstanding I think with renaming Common to Intercom, in this same PR the channel can also be made intrinsic to intercom assemblies as suggested by VoidGremlin, it's been brought up that HoP players struggle to scrounge up Common keys for people assembling intercoms, and instead of adding ROCommon/Intercom keys to the PTech it should just come for free with every intercom.

imatsoup:
> This is not the right way to do this. The loc string for the channel name should just be `Intercom`, and the _examine text_ should state that it's read only. Ergo, instead of `{$key} for {$name}` it should be `Read only: {$name}`. The current implementation makes the label over twice as long as any other radio channel, which ends up taking up valuable real estate in the chat box. The examine system should be able to check for the same NoHeadset bool on the channel prototype trivially.

Currently trying to work on implementing this idea for the examine text, but rn the way the examine text is implemented in EncryptionKeySystem makes it kind of difficult. Currently it has a pattern that it applies to every channel available in the headset/device, and I'm struggling to figure out how to make it apply a different pattern for one specific channel without overwriting the others.

imatsoup:
If I could apply an operation in the .ftl string itself that just checks for NoHeadset = true and uses the alternative pattern, that would be the most straightforward thing, but I'm not sure if that's possible. Gonna ask in the devbus today when I get a second.

imatsoup:
> Frankly im unsure if we should move to make this change and continue work on this now to the future. popular opinion is against the common channels and as much as id love to keep it like this it might be better to plan for a revert, address the issues that are present in the system and give alternative methods of communication, then plan a more worked out re-implementation.

I dont think there's any harm in continuing to improve it while things are like this. If it turns out to really be that we can't make people happy with it no matter what, its as easy as commenting out or removing a field from the common key and reverting this PR. I'm not certain about it - I know there have been complaints but actually playing the round I still feel strongly that its an improvement over the old system.

Resources/Locale/en-US/radio/components/encryption-key-component.ftl
```diff
@@ -6,5 +6,10 @@ encryption-keys-are-locked = Encryption key slots are locked!
 encryption-keys-panel-locked = Open maintenance panel first!
 
 examine-encryption-channels-prefix = Available frequencies:
-examine-encryption-channel = [color={$color}]{$key} for {$id} ({NATURALFIXED($freq, 1)})[/color]
+# Carpmosia-start - Intercom to Common
+examine-encryption-channel = { $noHeadset ->
+[true] [color={$color}]:read-only {$id} ({NATURALFIXED($freq, 1)}) [/color]
```
arenaconspiracy:
```suggestion
[true] [color={$color}]read-only {$id} ({NATURALFIXED($freq, 1)}) [/color]
```
doesn't need the colon because that's meant to showcase a digraph in messages `:s bro what`

Resources/Prototypes/Entities/Objects/Devices/encryption_keys.yml
arenaconspiracy: I think intercom units should be given the channel intrinsically and the description can stay, while still _technically false_ in practice there wouldn't be a way to disprove that ingame with current systems.

arenaconspiracy (approved):
NOW it does the right thing, but I think with these changes the PR can bundle intercom units having the Intercom channel intrinsically, bypassing the need for a spare key when building new ones.

Resources/Locale/en-US/radio/components/encryption-key-component.ftl
imatsoup: Honestly when I saw it in game without the : it looked weird to my eyes. That's pretty much the only reason I added it there. You are correct however

Resources/Prototypes/Entities/Objects/Devices/encryption_keys.yml
imatsoup: Probably right. I messed around with trying to have it be that way but couldn't get it to work in my own time.

imatsoup:
Updated PR description and changelog

Resources/Locale/en-US/radio/components/encryption-key-component.ftl
```diff
@@ -6,5 +6,10 @@ encryption-keys-are-locked = Encryption key slots are locked!
 encryption-keys-panel-locked = Open maintenance panel first!
 
 examine-encryption-channels-prefix = Available frequencies:
-examine-encryption-channel = [color={$color}]{$key} for {$id} ({NATURALFIXED($freq, 1)})[/color]
+# Carpmosia-start - Intercom to Common
+examine-encryption-channel = { $noHeadset ->
+[true] [color={$color}]read-only {$id} ({NATURALFIXED($freq, 1)}) [/color]
+*[other] [color={$color}]{$key} for {$id} ({NATURALFIXED($freq, 1)})[/color]
+}
+# Carpmosia-end - Intercom to Common
```
arenaconspiracy:
```suggestion
# Carpmosia-start - Common to Intercom
examine-encryption-channel = { $noHeadset ->
[true] [color={$color}]read-only {$id} ({NATURALFIXED($freq, 1)}) [/color]
*[other] [color={$color}]{$key} for {$id} ({NATURALFIXED($freq, 1)})[/color]
}
# Carpmosia-end - Common to Intercom
```

arenaconspiracy (commented):
I think the comments from #272 should also be updated to the same feature name used in this.

Resources/Locale/en-US/radio/components/encryption-key-component.ftl
imatsoup: woops

arenaconspiracy (approved):
Looks good to me and in my opinion additions are either out of scope or frivolous. I think in the interest of time and player experience it's imperative this is merged as soon as possible.

Resources/Locale/en-US/radio/components/encryption-key-component.ftl
```diff
@@ -6,5 +6,10 @@ encryption-keys-are-locked = Encryption key slots are locked!
 encryption-keys-panel-locked = Open maintenance panel first!
 
 examine-encryption-channels-prefix = Available frequencies:
-examine-encryption-channel = [color={$color}]{$key} for {$id} ({NATURALFIXED($freq, 1)})[/color]
+# Carpmosia-start - Common to Intercom
+examine-encryption-channel = { $noHeadset ->
+[true] [color={$color}]read-only {$id} ({NATURALFIXED($freq, 1)}) [/color]
+*[other] [color={$color}]{$key} for {$id} ({NATURALFIXED($freq, 1)})[/color]
+}
+# Carpmosia-end - Common to Intercom
```
SnappingOpossum: Double hashes are regions in ftl.
```suggestion
## Carpmosia-start - Common to Intercom
examine-encryption-channel = { $noHeadset ->
[true] [color={$color}]read-only {$id} ({NATURALFIXED($freq, 1)}) [/color]
*[other] [color={$color}]{$key} for {$id} ({NATURALFIXED($freq, 1)})[/color]
}
## Carpmosia-end - Common to Intercom
```

SnappingOpossum (approved):

imatsoup:
Added DNM tag due to https://github.com/carpmosia/carpmosia/pull/294

CarpBot (changes requested):

PR #291, opened 2026-03-19
Revert workflow fix attempt

Atakku:
## General information

Reverted workflow tweaks that were caused by me going insane as Github's workflows were shitting themselves (they were just dying randomly during test)

## Why / Balance

**work**flows as the name implies are supposed to **work**


CarpBot (changes requested):

PR #294, opened 2026-03-21
Revert #272

Atakku:
## General information

Reverted #272 as the experiment has concluded

## Why / Balance

Experiment concluded

## Changelog

:cl:
- tweak: Common speaking removal experiment concluded, and has been reverted for the time being.


Resources/Prototypes/radio_channels.yml
```diff
@@ -4,7 +4,6 @@
   keycode: ";"
   frequency: 1459
   color: "#2cdb2c"
-  noHeadset: true # Carpmosia-edit - no-headset channels
```
arenaconspiracy, line 7: You don't need to revert the feature wholesale, you can axe this one line to return it to previous behavior. This means re-introduction down the line can be made trivial and can be rolled into e.g. #290 paving the road for the future.

arenaconspiracy (changes requested):
See file comment, full reversion is unnecessary and keeping support opens opportunities for further PRs if the intent is to reinstate #272 in the future.

CarpBot (changes requested):

Resources/Prototypes/radio_channels.yml
Atakku: i can just revert the revert later when needed
its kinda implemented in a mediocre way anyway rn

PR #295, opened 2026-03-22
Fixed Admeme Pneumatic Cannon

Timtee:
## General information
The Admeme Pneumatic Cannon had a bad whitelist remaining from however long ago- thus making it nonfunctional previously. With this PR, it should fix that for good.

## Why / Balance
This should allow admins to utilize it again to throw all sorts of things at stuff.

## Technical details
Simply deleted the whitelist portion from the Storage component area for the admeme pneumatic cannon as well as the part of the PneumaticCannon component for not making items act as if they've been thrown, thus making such not automatically defined

## Media
I had made the change initially to a localized server of mine as a test, it being tested [here](https://youtu.be/m4rTzobQkU0)

## Changelog


CarpBot (changes requested):

Atakku (approved):
mrrp miau

PR #296, opened 2026-03-22
Attempt at fixing tests

Atakku:
## General information

Fixed the Build & Test Debug workflow dying with "The operation was canceled."

## Why / Balance

We need the workflows to work, otherwise nothing flows


CarpBot (changes requested):

Atakku:
I LOVE INCONSISTENCY!!!!!!!!!!

PR #303, opened 2026-03-30
Make whistles usable from the actions bar

arenaconspiracy:
## General information
Adds an item action to whistles, letting you use them hands-free. The delay between the action & use-in-hand is synchronized.
Re-implementation of https://github.com/RMC-14/RMC-14/pull/3838 , shimmed into the upstream files instead.

## Why / Balance
Whistles are a really cool & generally useful way of getting people's attention in a crowd during a heated moment, but get a lot of friction _in heated moments_ by making you pull them off your neck into a free hand at speed. Making them use an action is a proven true method of making them much more accessible to use in practice.
Making whistles more prevalent opens the door to making more kinds of unique whistles in the future, a la the Head of Security having an e-whistle with a distinct sound to command their department.

## Technical details
Adds a simple EmitSoundOnActionComponent that's as nothingburger as every EmitSoundOn[X]Comp.
Adds a simple action prototype & defines it in an action datafield in WhistleComponent.
Makes WhistleSystem subscribe to GetItemActions and the new SoundAction, skips out on adding the actions if the item is in a pocket.
Defines EmitSoundOnAction on the base and trench whistles.

## Media

https://github.com/user-attachments/assets/6f93afed-36e6-4c72-b38a-da3621f812ac

## Breaking changes
N/A

## Changelog
:cl:
- add: Whistles can now be used from the action bar while worn around your neck.


arenaconspiracy:
~~A known bug is the whistle actions letting you play the sound 2 or 3 times if you spam it, which I'm not sure the cause of or how to fix. Opening this for review so someone else might catch it & I think the bug is minor enough to hardly matter.~~ I fixed it. The action was being handled improperly.

CarpBot (changes requested):

imatsoup (approved):
I like it. Checkfail is unrelated, tests are just fucky right now

imatsoup (approved):
after the bugfix there is no reason to not have this as a feature now

SnappingOpossum (approved):
Tests run fine on a fixed branch.

Atakku:
wouldn't it be better to just have a generic use action event? that could automatically fire the onUse stuff?

Atakku:
although i suppose whistles cant be used while equipped on the neck slot

arenaconspiracy:
OnUse doesn't let you use it from the --> actions bar <-- which is the goal of making it hands-free.

Atakku:
> OnUse doesn't let you use it from the --> actions bar <-- which is the goal of making it hands-free. 

No what I meant is triggering onuse from action, so you don't need to redefine interactions

arenaconspiracy:
image
image
image

Not to mention that firing OnUse from an action is far from perfect or even a very good idea for what it's worth.

Atakku:
> Not to mention that firing OnUse from an action is far from perfect or even a very good idea for what it's worth.

ok whatever, i dont have time to impl it myself anyway

Atakku (approved):
whistlemaxxing

Image

CarpBot (approved):

PR #312, opened 2026-04-14
Upstream merge fixes

Atakku:
## General information

Fixed an upstream merge mistake

## Why / Balance

Fix


CarpBot (changes requested):

PR #311, opened 2026-04-12
Add pointing modifiers to items

arenaconspiracy:
Port & extension of https://github.com/impstation/imp-station-14/pull/3337
## General information
Adds support for items in your active hand modifying the popup phrase & icon of the pointing action.
Adds four new icons for sharp melee weapons (dagger icon), blunt melee weapons (sledgehammer icon), firearms (crosshair icon), and the microphone (microphone).
Adds two new phrases, `take[s] aim at`, `gesture[s] to`.

## Why / Balance
Letting you Visibly take aim with a gun or point a knife at someone paves the way for roleplay pizzazz where you can Very Clearly threaten someone at gunpoint/knifepoint. Letting you gesture with a microphone lets you Very Clearly ask someone to speak for an interview. Sledgehammer. Et cetera et cetera. The implementation allows for an arbitrary amount of icons and phrases be tacked on in the future for specific items, such as chainsaws or stun batons, that I haven't bothered with yet.

## Technical details
Registers a new `PointingModifier` component with fields for pointer `EntProtoId` & `LocId` phrases.
Modifies `PointingSystem` to get the item in the active hand, `TryComp`s for `PointingModifier` and uses it's fields for the pointer and phrases.
Folded the `Loc.GetString` calls in `PointingSystem` for readability, and added the phrase key. Modifies the upstream FTL with the locales.
Made several `Pointer[Icon]` entity prototypes for the pointers, and several `BasePointer[Icon]` abstracts to supply `PointingModifier`, added the latter as parents to items that need a modifier.
Keeps the upstream `PointingArrow` entity in an unused state, made a new functionally `PointerArrow` that follows the new naming scheme.
Adds `EntityCategoryAttribute` to `SharedPointingArrowComponent` to automatically mark all pointing arrows with `HideSpawnMenu`.

## Media
image

https://github.com/user-attachments/assets/8d2663a6-6051-4724-8abb-861aff48baec

Following images are NOT exhaustive representations of affected items.

image
image
image

## Breaking changes
N/A

## Changelog
:cl:
- add: Most melee weapons, firearms, and the microphone now modify the icon & phrase used for pointing while in your active hand.

PR #315, opened 2026-04-19
Actually fix comments of #311

SnappingOpossum:
## General information

Added back surround comments, removed a start comment with no end.

## Why / Balance

Carpmosia comment guidelines.


CarpBot (changes requested):

Atakku:
i removed the wrong comment :sob:

Atakku (approved):

PR #324, opened 2026-05-03
AOE system and testing-entities

Chickon20:
## General information

Area of effect system that introduces AOE which is bound to entities

## Why / Balance

A new system to apply damage to multiple things without making an explosion was kind of already needed.

## Technical details

- added AreaOfEffectComponent.cs
- added AreaOfEffectSystem.cs
- added test_aoe_item.yml for testing purposes
- added test_aoe_structure.yml for testing purposes
## Breaking changes

- no issues for the moment

### DIFF
--- a/Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectComponent.cs
+++ b/Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectComponent.cs
@@ -1,0 +1,44 @@
+using Content.Shared.Damage.Prototypes;
+using Content.Shared.FixedPoint;
+using Content.Shared.Whitelist;
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared._Carpmosia.AreaOfEffect;
+
+/// <summary>
+/// This is used for...
+/// </summary>
+[RegisterComponent]
+[AutoGenerateComponentState]
+public sealed partial class AreaOfEffectComponent : Component
+{
+    [DataField]
+    public Dictionary<ProtoId<DamageTypePrototype>, FixedPoint2> Damage = new();
+
+    [DataField]
+    public float Cooldown = 1f;
+
+    /// <summary>
+    /// The radius around the entity within which damage is applied.
+    /// </summary>
+    [DataField]
+    public float Radius = 5f;
+
+    /// <summary>
+    /// The time when the next damage application should occur.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public float NextApplicationTime;
+
+    /// <summary>
+    /// Whitelist of entities to damage. If set, only entities matching this whitelist will be damaged.
+    /// </summary>
+    [DataField]
+    public EntityWhitelist? Whitelist;
+
+    /// <summary>
+    /// Blacklist of entities to exclude from damage. If set, entities matching this blacklist will not be damaged.
+    /// </summary>
+    [DataField]
+    public EntityWhitelist? Blacklist;
+}
--- a/Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectComponent.cs
+++ b/Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectComponent.cs
@@ -28,7 +28,7 @@ public sealed partial class AreaOfEffectComponent : Component
     /// The time when the next damage application should occur.
     /// </summary>
     [DataField, AutoNetworkedField]
-    public float NextApplicationTime;
+    public TimeSpan? NextApplicationTime;
 
     /// <summary>
     /// Whitelist of entities to damage. If set, only entities matching this whitelist will be damaged.
--- a/Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectComponent.cs
+++ b/Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectComponent.cs
@@ -1,7 +1,5 @@
-using Content.Shared.Damage.Prototypes;
-using Content.Shared.FixedPoint;
+using Content.Shared.Damage;
 using Content.Shared.Whitelist;
-using Robust.Shared.Prototypes;
 
 namespace Content.Shared._Carpmosia.AreaOfEffect;
 
@@ -24,7 +22,7 @@ public sealed partial class AreaOfEffectComponent : Component
      /// The damage to apply to entities within the AoE. The key is the damage type, and the value is the amount of damage to apply.
      /// </summary>
     [DataField]
-    public Dictionary<ProtoId<DamageTypePrototype>, FixedPoint2> Damage = new();
+    public DamageSpecifier Damage = new();
 
     /// <summary>
     /// Cooldown for how much time has to pass before AoE is applied again.
@@ -37,6 +35,12 @@ public sealed partial class AreaOfEffectComponent : Component
     /// </summary>
     [DataField]
     public TimeSpan? Duration;
+
+    /// <summary>
+    /// Determines if damage should be spread between entities.
+    /// </summary>
+    [DataField]
+    public bool DamageSpread =  true;
 
     /// <summary>
     /// Whitelist of entities to damage. If set, only entities matching this whitelist will be damaged.
--- a/Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectComponent.cs
+++ b/Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectComponent.cs
@@ -16,7 +16,7 @@ public sealed partial class AreaOfEffectComponent : Component
     public Dictionary<ProtoId<DamageTypePrototype>, FixedPoint2> Damage = new();
 
     [DataField]
-    public float Cooldown = 1f;
+    public TimeSpan Cooldown = TimeSpan.FromZero;
 
     /// <summary>
     /// The radius, in tiles, around the entity within which damage is applied.
--- a/Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectComponent.cs
+++ b/Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectComponent.cs
@@ -6,29 +6,37 @@ using Robust.Shared.Prototypes;
 namespace Content.Shared._Carpmosia.AreaOfEffect;
 
 /// <summary>
-/// This is used for...
+/// Component that defines an area of effect (AoE) damage dealer.
+/// The AoE damage dealer applies damage to entities within a certain radius,
+/// optionally whitelisting or blacklisting certain entities.
+/// It can also have a cooldown and a limited duration.
 /// </summary>
 [RegisterComponent]
-[AutoGenerateComponentState]
 public sealed partial class AreaOfEffectComponent : Component
 {
+     /// <summary>
+     /// The radius, in tiles, around the entity within which damage is applied.
+     /// </summary>
+     [DataField]
+     public float Radius = 5f;
+
+     /// <summary>
+     /// The damage to apply to entities within the AoE. The key is the damage type, and the value is the amount of damage to apply.
+     /// </summary>
     [DataField]
     public Dictionary<ProtoId<DamageTypePrototype>, FixedPoint2> Damage = new();
 
-    [DataField]
-    public TimeSpan Cooldown = TimeSpan.FromZero;
-
     /// <summary>
-    /// The radius, in tiles, around the entity within which damage is applied.
+    /// Cooldown for how much time has to pass before AoE is applied again.
     /// </summary>
     [DataField]
-    public float Radius = 5f;
+    public TimeSpan Cooldown = TimeSpan.FromSeconds(1);
 
     /// <summary>
-    /// The time when the next damage application should occur.
+    /// Optional duration for how long the area of effect lasts. If set, the component will be automatically removed after this time.
     /// </summary>
-    [DataField, AutoNetworkedField]
-    public TimeSpan? NextApplicationTime;
+    [DataField]
+    public TimeSpan? Duration;
 
     /// <summary>
     /// Whitelist of entities to damage. If set, only entities matching this whitelist will be damaged.
--- a/Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectComponent.cs
+++ b/Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectComponent.cs
@@ -19,7 +19,7 @@ public sealed partial class AreaOfEffectComponent : Component
     public float Cooldown = 1f;
 
     /// <summary>
-    /// The radius around the entity within which damage is applied.
+    /// The radius, in tiles, around the entity within which damage is applied.
     /// </summary>
     [DataField]
     public float Radius = 5f;
--- a/Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectSystem.cs
+++ b/Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectSystem.cs
@@ -76,8 +76,7 @@ public sealed class AreaOfEffectSystem : EntitySystem
     {
         var result = new List<EntityUid>();
 
-        if (!TryComp(uid, out TransformComponent? aoeTransform))
-            return result;
+        var aoeTransform = Transform(uid);
 
         if (!TryComp(uid, out AreaOfEffectComponent? aoe))
             return result;
--- a/Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectSystem.cs
+++ b/Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectSystem.cs
@@ -93,12 +93,6 @@ public sealed class AreaOfEffectSystem : EntitySystem
     /// <param name="aoe">The area of effect component</param>
     private void ApplyAreaOfEffectDamage(EntityUid uid, AreaOfEffectComponent aoe)
     {
-        var aoeTransform = Transform(uid);
-        var aoePosition = _transform.GetWorldPosition(aoeTransform);
-
-        _entitiesInRange.Clear();
-        _lookup.GetEntitiesInRange(aoeTransform.MapID, aoePosition, aoe.Radius, _entitiesInRange);
-
         if (!_damageSpecifiers.TryGetValue(uid, out var damageSpecifier))
         {
             damageSpecifier = new DamageSpecifier();
@@ -109,18 +103,14 @@ public sealed class AreaOfEffectSystem : EntitySystem
             _damageSpecifiers[uid] = damageSpecifier;
         }
 
-        foreach (var targetUid in _entitiesInRange)
+        foreach (var targetUid in _lookup.GetEntitiesInRange<DamageableComponent>(Transform(uid).Coordinates, aoe.Radius))
         {
-            // Filter for damageable entities
-            if (!TryComp(targetUid, out DamageableComponent? _))
-                continue;
-
             // Check whitelist/blacklist filtering
             if (!_whitelist.CheckBoth(targetUid, aoe.Blacklist, aoe.Whitelist))
                 continue;
 
             _damageable.TryChangeDamage(
-                (targetUid, null),
+                targetUid!,
                 damageSpecifier,
                 ignoreResistances: false,
                 interruptsDoAfters: true,
--- a/Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectSystem.cs
+++ b/Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectSystem.cs
@@ -33,7 +33,7 @@ public sealed class AreaOfEffectSystem : EntitySystem
         _timings[uid] = new AreaOfEffectTiming
         {
             NextApplicationTime = null,
-            StartTime = _timing.CurTime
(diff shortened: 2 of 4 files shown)


Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectComponent.cs
```diff
@@ -0,0 +1,44 @@
+using Content.Shared.Damage.Prototypes;
+using Content.Shared.FixedPoint;
+using Content.Shared.Whitelist;
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared._Carpmosia.AreaOfEffect;
+
+/// <summary>
+/// This is used for...
```
arenaconspiracy: Fill out the placeholder.

CarpBot (dismissed):

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectComponent.cs
```diff
@@ -0,0 +1,44 @@
...
+/// </summary>
+[RegisterComponent]
+[AutoGenerateComponentState]
+public sealed partial class AreaOfEffectComponent : Component
+{
+    [DataField]
+    public Dictionary<ProtoId<DamageTypePrototype>, FixedPoint2> Damage = new();
+
+    [DataField]
+    public float Cooldown = 1f;
```
arenaconspiracy: Do Not Use Floats For Timing Or You Will Go To Hell Before You Die.
Tack on `using Robust.Shared.Timing` and make this a TimeSpan
```suggestion
    public TimeSpan Cooldown = TimeSpan.FromZero;
```

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectComponent.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    /// <summary>
+    /// The radius around the entity within which damage is applied.
```
arenaconspiracy: Nitpick.
```suggestion
    /// The radius, in tiles, around the entity within which damage is applied.
```

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectComponent.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// </summary>
+    [DataField]
+    public float Radius = 5f;
+
+    /// <summary>
+    /// The time when the next damage application should occur.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public float NextApplicationTime;
```
arenaconspiracy: See above. This also could maybe maybe maybe stand to be nullable.
```suggestion
    public TimeSpan? NextApplicationTime;
```

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectSystem.cs
```diff
@@ -0,0 +1,114 @@
...
+    [Dependency] private readonly DamageableSystem _damageable = default!;
+    [Dependency] private readonly IGameTiming _timing = default!;
+    [Dependency] private readonly EntityWhitelistSystem _whitelist = default!;
+
+    /// <inheritdoc/>
+    public override void Update(float frameTime)
+    {
+        base.Update(frameTime);
+
+        var query = EntityQueryEnumerator<AreaOfEffectComponent>();
```
arenaconspiracy, line 50: You should use dependency query injection instead.

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectComponent.cs
```diff
@@ -0,0 +1,44 @@
...
+
+/// <summary>
+/// This is used for...
+/// </summary>
+[RegisterComponent]
+[AutoGenerateComponentState]
+public sealed partial class AreaOfEffectComponent : Component
+{
+    [DataField]
+    public Dictionary<ProtoId<DamageTypePrototype>, FixedPoint2> Damage = new();
```
SnappingOpossum: Use a DamageSpecifier instead.

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectSystem.cs
```diff
@@ -0,0 +1,114 @@
...
+/// This handles area of effect damage application to nearby damageable entities.
+/// </summary>
+public sealed class AreaOfEffectSystem : EntitySystem
+{
+    [Dependency] private readonly SharedTransformSystem _transform = default!;
+    [Dependency] private readonly DamageableSystem _damageable = default!;
+    [Dependency] private readonly IGameTiming _timing = default!;
+    [Dependency] private readonly EntityWhitelistSystem _whitelist = default!;
+
+    /// <inheritdoc/>
```
arenaconspiracy, line 44: Erroneous tag.

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectComponent.cs
SnappingOpossum: Times should use Timespans.

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectComponent.cs
SnappingOpossum: See above about timespans, you can use a TimeOffsetSerializer for this one.

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectSystem.cs
```diff
@@ -0,0 +1,114 @@
...
+    public override void Update(float frameTime)
```
arenaconspiracy, line 45: This is going to absolutely obliterate performance in it's current state and you should at the least put the cooldown checks at the start of the function. I would provide an example but I'm currently on mobile and can neither pull up existing uses or write it off the top of my head. Stand by.

Resources/Prototypes/_Carpmosia/Entities/Objects/Misc/test_aoe_item.yml
```diff
@@ -0,0 +1,18 @@
+- type: entity
+  id: TestAreaOfEffectItem
+  parent: BaseItem
```
SnappingOpossum: Parent should be above id.

Resources/Prototypes/_Carpmosia/Entities/Structures/test_aoe_structure.yml
```diff
@@ -0,0 +1,18 @@
+- type: entity
+  id: TestAreaOfEffectStructure
+  parent: BaseStructure
```
SnappingOpossum: ^

SnappingOpossum (changes requested):

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectSystem.cs
```diff
@@ -0,0 +1,114 @@
...
+            return result;
+
+        if (!TryComp(uid, out AreaOfEffectComponent? aoe))
+            return result;
+
+        var aoePosition = _transform.GetWorldPosition(aoeTransform);
+        var aoeMap = aoeTransform.MapID;
+
+        // Query all damageable entities
+        var query = EntityQueryEnumerator<DamageableComponent>();
```
arenaconspiracy: See above.

Resources/Prototypes/_Carpmosia/Entities/Objects/Misc/test_aoe_item.yml
```diff
@@ -0,0 +1,18 @@
...
+  name: test AoE damage item
+  description: A test item that applies area of effect damage periodically. For testing purposes only.
```
arenaconspiracy: Admins will not read this and proceed to slop around with it. It's better if it's childproofed.
```suggestion
  categories: [ HideSpawnMenu ]
  suffix: Debug
```

arenaconspiracy (changes requested):
It's bad practice to add a system without adding it's uses at the same time. Nobody wants unused code lying around. Add the actual end goal use for the system once you're done with it.

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectSystem.cs
```diff
@@ -0,0 +1,114 @@
...
+    /// </summary>
+    /// <param name="uid">The area of effect entity</param>
+    /// <param name="radius">The detection radius</param>
+    /// <returns>A list of damageable entities within the radius</returns>
+    public List<EntityUid> GetNearbyDamageableEntities(EntityUid uid, float radius)
+    {
+        var result = new List<EntityUid>();
+
+        if (!TryComp(uid, out TransformComponent? aoeTransform))
+            return result;
```
SnappingOpossum: This can just changed to this:
```suggestion
        var aoeTransform = Transform(uid);
```
Since all entities have a transform

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectSystem.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        if (!TryComp(uid, out AreaOfEffectComponent? aoe))
+            return result;
```
SnappingOpossum: Pass in the AreaOfEffectComponent instead.

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectSystem.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        var aoePosition = _transform.GetWorldPosition(aoeTransform);
+        var aoeMap = aoeTransform.MapID;
+
+        // Query all damageable entities
+        var query = EntityQueryEnumerator<DamageableComponent>();
+        while (query.MoveNext(out var damageableUid, out _))
+        {
+            if (!TryComp(damageableUid, out TransformComponent? damageableTransform))
+                continue;
```
SnappingOpossum: See above on transform.

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectSystem.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            // Check whitelist/blacklist filtering
+            if (!_whitelist.CheckBoth(damageableUid, aoe.Blacklist, aoe.Whitelist))
+                continue;
+
+            result.Add(damageableUid);
+        }
+
+        return result;
```
SnappingOpossum: Consider using a `GetEntitiesInRange<T>` and then checking the whitelist/blacklist instead of all this.

SnappingOpossum (changes requested):
Second pass

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectSystem.cs
Chickon20: thanks I didn't know there was already a method for that

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectSystem.cs
```diff
@@ -0,0 +1,130 @@
...
+            _timings[uid] = timing;
+        }
+    }
+
+    /// <summary>
+    /// Applies damage to all entities within the given radius.
+    /// </summary>
+    /// <param name="uid">The area of effect entity</param>
+    /// <param name="aoe">The area of effect component</param>
+    private void ApplyAreaOfEffectDamage(EntityUid uid, AreaOfEffectComponent aoe)
```
SnappingOpossum: You should just be passing this as an `Entity<AreaOfEffectComponent>`

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectSystem.cs
```diff
@@ -0,0 +1,130 @@
...
+            }
+
+            var nextTime = _timings[uid].NextApplicationTime;
+
+            // Check if it's time to apply damage
+            if (nextTime.HasValue && nextTime > curTime)
+                continue;
+
+            // Apply damage to nearby entities
+            ApplyAreaOfEffectDamage(uid, aoe);
```
SnappingOpossum: For the part below you should be able to do this to convert it to an `Entity<T>`
```suggestion
            ApplyAreaOfEffectDamage((uid, aoe));
```

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectComponent.cs
```diff
@@ -0,0 +1,52 @@
...
+     /// The radius, in tiles, around the entity within which damage is applied.
+     /// </summary>
+     [DataField]
+     public float Radius = 5f;
+
+     /// <summary>
+     /// The damage to apply to entities within the AoE. The key is the damage type, and the value is the amount of damage to apply.
+     /// </summary>
+    [DataField]
+    public Dictionary<ProtoId<DamageTypePrototype>, FixedPoint2> Damage = new();
```
SnappingOpossum: This should still be a DamageSpecifier

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectSystem.cs
```diff
@@ -0,0 +1,130 @@
...
+        SubscribeLocalEvent<AreaOfEffectComponent, ComponentStartup>(OnStartup);
+        SubscribeLocalEvent<AreaOfEffectComponent, ComponentShutdown>(OnShutdown);
+    }
+
+    private void OnStartup(EntityUid uid, AreaOfEffectComponent component, ComponentStartup args)
+    {
+        _timings[uid] = new AreaOfEffectTiming
+        {
+            NextApplicationTime = null,
+            StartTime = _timing.CurTime
```
SnappingOpossum:
```suggestion
            StartTime = _timing.CurTime,
```

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectSystem.cs
```diff
@@ -0,0 +1,130 @@
...
+                continue;
+
+            _damageable.TryChangeDamage(
+                (targetUid, null),
+                damageSpecifier,
+                ignoreResistances: false,
+                interruptsDoAfters: true,
+                origin: uid);
+        }
+    }
```
SnappingOpossum, line 117: You can do something like this for the range lookup. No point in storing it if you're clearing it every time this is run.
```suggestion
    {
        if (!_damageSpecifiers.TryGetValue(uid, out var damageSpecifier))
        {
            damageSpecifier = new DamageSpecifier();
            foreach (var (damageTypeId, amount) in aoe.Damage)
            {
                damageSpecifier.DamageDict[damageTypeId] = amount;
            }
            _damageSpecifiers[uid] = damageSpecifier;
        }

        foreach (var targetUid in _lookup.GetEntitiesInRange<DamageableComponent>(Transform(uid).Coordinates, aoe.Radius))
        {
            // Check whitelist/blacklist filtering
            if (!_whitelist.CheckBoth(targetUid, aoe.Blacklist, aoe.Whitelist))
                continue;

            _damageable.TryChangeDamage(
                targetUid!,
                damageSpecifier,
                ignoreResistances: false,
                interruptsDoAfters: true,
                origin: uid);
        }
    }
```

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectSystem.cs
```diff
@@ -0,0 +1,130 @@
...
+{
+    [Dependency] private readonly SharedTransformSystem _transform = default!;
+    [Dependency] private readonly DamageableSystem _damageable = default!;
+    [Dependency] private readonly IGameTiming _timing = default!;
+    [Dependency] private readonly EntityWhitelistSystem _whitelist = default!;
+    [Dependency] private readonly EntityLookupSystem _lookup = default!;
+
+    private readonly HashSet<EntityUid> _entitiesInRange = new();
+    private readonly Dictionary<EntityUid, AreaOfEffectTiming> _timings = new();
+    private readonly Dictionary<EntityUid, DamageSpecifier> _damageSpecifiers = new();
```
SnappingOpossum: When you make the field a DamageSpecifier you should also remove this since it'll be on the component.

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectSystem.cs
```diff
@@ -0,0 +1,130 @@
...
+    {
+        _timings.Remove(uid);
+        _damageSpecifiers.Remove(uid);
+    }
+
+    private struct AreaOfEffectTiming
+    {
+        public TimeSpan? NextApplicationTime;
+        public TimeSpan StartTime;
+    }
```
SnappingOpossum: I don't see how this is any better than storing the values on the component. Also, these methods should use the `Entity<T>` format.

SnappingOpossum (dismissed):

imatsoup:
Hey there. Just a quick question after reading your PR description: Why do you not want to use an explosion for your breaching charge?

Chickon20:
> Hey there. Just a quick question after reading your PR description: Why do you not want to use an explosion for your breaching charge?

I think the charge should be slow and quiet, unlike explosions, which are supposed to be fast, violent and loud.

SnappingOpossum:
> > Hey there. Just a quick question after reading your PR description: Why do you not want to use an explosion for your breaching charge?
> 
> I think the charge should be slow and quiet, unlike explosions, which are supposed to be fast, violent and loud.

I don't see how this relates to doing damage in an area, especially considering destruction but itself is decently loud. The quieter option has always been deconstruction.

Chickon20:
> > > Hey there. Just a quick question after reading your PR description: Why do you not want to use an explosion for your breaching charge?
> > 
> > I think the charge should be slow and quiet, unlike explosions, which are supposed to be fast, violent and loud.
> 
> I don't see how this relates to doing damage in an area, especially considering destruction but itself is decently loud. The quieter option has always been deconstruction.

I think it makes sense for the charge to damage walls around it.
But since there is no way to apply AoE via yml I decided to make my own component.
I also didn't want to hardcode the charge in case someone wanted to do something similar.
Perhaps I can add an optional damage-spread to not make it too OP.

arenaconspiracy:
If I wanted a breaching charge I would make a dedicated BreachingCharge[Component|System] that depends on StickyComponent to work, and in an update loop applies damage to the entity it's stuck to.

There is no reason for a breaching charge to deal damage to multiple walls.

Over Genericising Is Bad, Actually. If you need a dedicated system for something _make it a dedicated system_. and only Don't Repeat Yourself when it actually matters.

Content.Shared/_Carpmosia/AreaOfEffect/AreaOfEffectSystem.cs
```diff
@@ -0,0 +1,118 @@
...
+
+    private void OnStartup(Entity<AreaOfEffectComponent> entity)
+    {
+        _timings[entity.Owner] = new AreaOfEffectTiming(_timing);
+    }
+
+    private void OnShutdown(Entity<AreaOfEffectComponent> entity)
+    {
+        _timings.Remove(entity.Owner);
+    }
```
SnappingOpossum, line 36: You can just do this:
```suggestion
    public override void Initialize()
    {
        base.Initialize();
        SubscribeLocalEvent<AreaOfEffectComponent, ComponentStartup>(OnStartup);
        SubscribeLocalEvent<AreaOfEffectComponent, ComponentShutdown>(OnShutdown);
    }

    private void OnStartup(Entity<AreaOfEffectComponent> entity, ref ComponentStartup args)
    {
        _timings[entity.Owner] = new AreaOfEffectTiming(_timing);
    }

    private void OnShutdown(Entity<AreaOfEffectComponent> entity, ref ComponentShutdown args)
    {
        _timings.Remove(entity.Owner);
    }
```

SnappingOpossum:
I'll just give some more personal thoughts here, I don't like the DamageSpread option, I feel it makes this much more awkward, especially compared to what something like radiation does, and I don't like the concept of the damage being split like so. For the component itself, I don't really see any good use cases for this, I think a breaching charge is better handled by being explosive of for the thermite route being something sticky that damages what it's on.

Chickon20:
> I'll just give some more personal thoughts here, I don't like the DamageSpread option, I feel it makes this much more awkward, especially compared to what something like radiation does, and I don't like the concept of the damage being split like so. For the component itself, I don't really see any good use cases for this, I think a breaching charge is better handled by being explosive of for the thermite route being something sticky that damages what it's on.

I can make it default to false if you want to.

wavesneglected:
@Chickon20 

Will you be continuing work and discussion on this System?

Chickon20:
> @Chickon20 
> 
> Will you be continuing work and discussion on this System?

Probably, I should think about some use cases and some other features I want to add to the system tho.

imatsoup:
closing this as part of cleaning up the backlog given this isn't something we really want rn

PR #286, opened 2026-03-16
Change body friction to apply differently based on crit/dead status

ultrafiche:
## General information

Makes it so the friction modifier introduced in https://github.com/carpmosia/carpmosia/pull/270 has different values for crit vs dead bodies. Both values have been tuned down a little.

## Why / Balance
Based on playing in-game and player feedback. Bodies are slightly less harder to move now.

## Technical details

The Incapacitated datafield is now named Incapacitation and is a MobState. Whenever friction is applied, it's read to choose the correct friction modifier to apply based on CritFrictionMod and DeadFrictionMod, which together used to be LimpFrictionMod.

Both of these values are VERY fiddly and should MAYBE be changed in the future maybe question mark ?
## Media

https://github.com/user-attachments/assets/a47c5c88-ffaf-4f45-a3cd-2a268f5cdcac

i encoded this wrong so its the shittiest looking thing imaginable but you get it

## Breaking changes
changes to datafields above dot dot dot

## Changelog

:cl: microfiche
- tweak: Bodies are now slightly less cumbersome to move.

PR #333, opened 2026-05-09
Power cautery

arenaconspiracy:
## General information
Implements the power cautery & adds it to the paramedic locker.

The power cautery functions essentially as a battery powered tourniquet, stemming the same amount of bleeding in the same amount of time.
The power cautery has a capacity of 5 uses on a full charge, and begins self-recharging 10 seconds after use.

## Why / Balance
Required for something else on the horizon.
Paramedics deserve a new, if minor utility, and an Infinite Tourniquet fits that role perfectly.
This helps to put an emphasis on paramedics in line with with #270 and #286

## Technical details
A new PowerCauterySystem with simple logic handling only bleeding & power draw because bolting the logic onto HealingSystem would be a remarkably bad idea and I'm not stupid.
Currently uses PowerCellVisuals because I couldn't be bothered to write a new visualizer and over time realized that there's not enough sprite real estate to bother with all 5 battery indicator steps.

## Media
image
image

https://github.com/user-attachments/assets/11724f36-c6e2-406b-9008-9b117506a772

## Breaking changes
N/A

## Changelog
:cl:
- add: The paramedic locker now starts with the power cautery, a bleeding stopper with a self recharging battery.


CarpBot (dismissed):
Awaiting approval from @GlupiyBrush, @Atakku, @imatsoup

GlupiyBrush:
why does it self recharge?

arenaconspiracy:
Medbay doesn't get universal rechargers mapped (on purpose for good reason) so paramedics wouldn't be able to recharge their own tool in their own department, and making them work on power cells would be a verisimilitude hell dimension.
It's a minor enough utility that I don't consider it self-recharging anywhere in the world to be a balance concern.
This is designed explicitly as an Infinite Use Tourniquet which stack up to 5, and I didn't want to raise the maximum uses above 5 if it recharges after a few seconds, and I didn't want to make it not self-recharge.

Content.Shared/_Carpmosia/Healing/Systems/PowerCauterySystem.cs
```diff
@@ -0,0 +1,233 @@
...
+    private bool IsBleeding(Entity<PowerCauteryComponent> cautery, Entity<DamageableComponent> target)
+    {
+        if (!TryComp<BloodstreamComponent>(target, out var bloodstream))
+            return false;
+
+        if (cautery.Comp.BloodlossModifier < 0 && bloodstream.BleedAmount > 0)
+            return true;
+
+        return false;
+    }
```
SnappingOpossum: Why does this even need to check if the cautery has a negative bloodloss modifier?

Content.Shared/_Carpmosia/Healing/Systems/PowerCauterySystem.cs
```diff
@@ -0,0 +1,233 @@
...
+        if (couldDraw)
+            _battery.UseCharge(ent, draw);
+
+        return couldDraw;
+    }
+
+    private bool MatchDamageContainers(List<ProtoId<DamageContainerPrototype>> containers, Entity<DamageableComponent?> target)
+    {
+        if (!Resolve(target, ref target.Comp, false))
+            return false;
```
SnappingOpossum: The component has already been resolved by TryCauterize, you should be able to make a non null version instead of having to resolve it again.

Content.Shared/_Carpmosia/Healing/Systems/PowerCauterySystem.cs
```diff
@@ -0,0 +1,233 @@
...
+
+        if (target.Comp.DamageContainerID is not null &&
+            !containers.Contains(target.Comp.DamageContainerID.Value))
```
SnappingOpossum: Should this check like this?
```suggestion
        if (target.Comp.DamageContainerID is null ||
            containers.Contains(target.Comp.DamageContainerID.Value))
```
Unless less you want to ensure it's not null on the target either, in which case you'd still not want to negate the contains though.

Content.Shared/_Carpmosia/Healing/Systems/PowerCauterySystem.cs
```diff
@@ -0,0 +1,233 @@
...
+
+    // cauterization
+    private bool TryCauterize(Entity<PowerCauteryComponent> cautery, Entity<DamageableComponent?> target, EntityUid user)
+    {
+        if (!Resolve(target, ref target.Comp, false))
+            return false;
+
+        // check if our container list matches our target
+        if (cautery.Comp.DamageContainers is not null &&
+            MatchDamageContainers(cautery.Comp.DamageContainers, target))
```
SnappingOpossum: Same as above:
```suggestion
        if (cautery.Comp.DamageContainers is null ||
            !MatchDamageContainers(cautery.Comp.DamageContainers, target))
```

Content.Shared/_Carpmosia/Healing/Systems/PowerCauterySystem.cs
```diff
@@ -0,0 +1,233 @@
...
+        {
+            _popupSystem.PopupClient(Loc.GetString("medical-item-cant-use", ("item", cautery.Owner)), cautery, user);
+            return false;
+        }
+
+        // start sound
+        _audio.PlayPredicted(cautery.Comp.BeginSound, cautery, user);
+
+        // let the target know we're helping
+        bool isNotSelf = user != target.Owner;
```
SnappingOpossum: Might be better to flip this bool so there's not the double negation later

SnappingOpossum (changes requested):
Test fails real.

I would like to see this also deal damage while stopping bloodloss, since it is still burning the wounds closed, and this might be able to benefit from being more generic in what it can consume instead of being a system solely for battery powered ones

GlupiyBrush:
what stops you from using cautery (i think???), since it also stops the bleeding???

arenaconspiracy:
I don't think the cautery needs to deal any damage because 1) tourniquets used to do damage on use and got laughed out of the building for it 2) it's a precision tool so verisimilitude should dictate it doesn't burn more than it should 3) I don't want to add healing support to the system

I deliberately made the doafter on yourself take twice as long unlike tourniquets to compensate for it's infinite nature. I don't think it's a problem since there'll only ever be 1 to 3 power cauters at a time and bleeding is Not That Serious Anyway.

arenaconspiracy:
> it also stops the bleeding???

I don't know what you mean, it _only_ stops bleeding.

arenaconspiracy:
Nevermind I realized what you mean after having coffee.
The current cautery in the game "cauterizes" wounds by way of dealing heat damage in melee. Which is neither convenient or effective, whacking someone with a lit welding tool twice is more effective than using the surgical cautery. Hence the creation of a discrete system for this.

imatsoup (approved):
conceptual approval. i did have the thought that this could go onto like, the paramedic borg module, but idk if that would be considered bloat

arenaconspiracy:
I looked over all the medical borg modules to see if it would go in any one of them and realized that borg modules are actually a huge mess and I don't want to fuck with them without scrapping them all and starting over.

GlupiyBrush (approved):

SnappingOpossum:
> I don't think the cautery needs to deal any damage because 1) tourniquets used to do damage on use and got laughed out of the building for it 2) it's a precision tool so verisimilitude should dictate it doesn't burn more than it should 3) I don't want to add healing support to the system

I mean a part of the reason tourniquets dealing damage was so bad was that it made them unable to stop bleeding, and I'm not expecting a significant amount of damage, just a little to show that it was burned shut.

Resources/Prototypes/_Carpmosia/Entities/Objects/Specific/Medical/healing.yml
```diff
@@ -10,7 +10,7 @@
   parent: BaseItem
   id: PowerCautery
   name: power cautery
-  description: A miniature energy projector that creates a flash of superheated plasma at the pull of a trigger. it's integrated microreactor begins recharging ten seconds after use to avoid burnout.
+  description: A miniature energy projector that creates a flash of superheated plasma at the pull of a trigger. It's integrated microreactor begins recharging ten seconds after use to avoid burnout.
```
SnappingOpossum: Oh right, "it's" is only "it is", this should be "Its"

Content.Shared/_Carpmosia/Healing/Systems/PowerCauterySystem.cs
```diff
@@ -0,0 +1,219 @@
...
+/// </remarks>
+public sealed class PowerCauterySystem : EntitySystem
+{
+    [Dependency] private readonly SharedAudioSystem _audio = default!;
+    [Dependency] private readonly ISharedAdminLogManager _adminLogger = default!;
+    [Dependency] private readonly SharedBatterySystem _battery = default!;
+    [Dependency] private readonly SharedBloodstreamSystem _bloodstreamSystem = default!;
+    [Dependency] private readonly SharedDoAfterSystem _doAfter = default!;
+    [Dependency] private readonly SharedInteractionSystem _interactionSystem = default!;
+    [Dependency] private readonly SharedPopupSystem _popupSystem = default!;
```
SnappingOpossum:
```suggestion
public sealed partial class PowerCauterySystem : EntitySystem
{
    [Dependency] private SharedAudioSystem _audio = default!;
    [Dependency] private ISharedAdminLogManager _adminLogger = default!;
    [Dependency] private SharedBatterySystem _battery = default!;
    [Dependency] private SharedBloodstreamSystem _bloodstreamSystem = default!;
    [Dependency] private SharedDoAfterSystem _doAfter = default!;
    [Dependency] private SharedInteractionSystem _interactionSystem = default!;
    [Dependency] private SharedPopupSystem _popupSystem = default!;
```

Content.Shared/_Carpmosia/Healing/Systems/PowerCauterySystem.cs
```diff
@@ -0,0 +1,219 @@
...
+using Robust.Shared.Audio.Systems;
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared.Healing.Systems;
+
+/// <summary>
+/// Modifies bleeding stacks per doafter, at the cost of power.
+/// </summary>
+/// <remarks>
+/// Separate from <see cref="HealingSystem"> because this wants ONLY bleeding and not damage,
```
SnappingOpossum: Not closed properly and it can't resolve the HealingSystem symbol because it isn't imported.
```suggestion
/// Separate from <see cref="HealingSystem"/> because this wants ONLY bleeding and not damage,
```

Content.Shared/_Carpmosia/Healing/Systems/PowerCauterySystem.cs
```diff
@@ -0,0 +1,219 @@
...
+    {
+        if (!TryComp<BatteryComponent>(ent, out var battery))
+            return;
+
+        var charges = (int)(_battery.GetCharge((ent, battery)) / ent.Comp.PowerDraw);
+        args.PushMarkup(Loc.GetString("power-cautery-charge-examine", ("charges", charges)));
+    }
+
+    // checks
+    private bool IsBleeding(Entity<PowerCauteryComponent> cautery, Entity<InjurableComponent> target)
```
SnappingOpossum: `Entity<PowerCauteryComponent> cautery` is not used here and can be removed.

Content.Shared/_Carpmosia/Healing/Systems/PowerCauterySystem.cs
```diff
@@ -0,0 +1,219 @@
...
+
+    private bool GetAndDrawPower(Entity<BatteryComponent?> ent, float draw)
+    {
+        bool couldDraw = (_battery.GetCharge(ent) >= draw);
+
+        if (couldDraw)
+            _battery.UseCharge(ent, draw);
+
+        return couldDraw;
+    }
```
SnappingOpossum, line 97: Should be able to replace this with TryUseCharge.

Content.Shared/_Carpmosia/Healing/Systems/PowerCauterySystem.cs
```diff
@@ -0,0 +1,219 @@
...
+
+    private bool MatchDamageContainers(List<ProtoId<DamageContainerPrototype>>? containers, InjurableComponent injurable)
```
SnappingOpossum: Can be made static.

Content.Shared/_Carpmosia/Healing/Systems/PowerCauterySystem.cs
```diff
@@ -0,0 +1,219 @@
...
+        if (!IsBleeding(cautery, target!))
+        {
+            _popupSystem.PopupClient(Loc.GetString("medical-item-cant-use", ("item", cautery.Owner)), cautery, user);
+            return false;
+        }
+
+        // start sound
+        _audio.PlayPredicted(cautery.Comp.BeginSound, cautery, user);
+
+        bool isSelf = user == target.Owner;
```
SnappingOpossum, line 135: Can use var instead of bool here.

Content.Shared/_Carpmosia/Healing/Systems/PowerCauterySystem.cs
```diff
@@ -0,0 +1,219 @@
...
+            return;
+
+        if (!TryComp(args.Used, out BatteryComponent? battery))
+            return;
+
+        // get target bloodstream
+        TryComp<BloodstreamComponent>(target, out var bloodstream);
+
+        // try to draw from our battery, fail if we can't
+        bool powered = GetAndDrawPower((args.Used.Value, battery), cautery.PowerDraw);
```
SnappingOpossum, line 175: See above.

Content.Shared/_Carpmosia/Healing/Systems/PowerCauterySystem.cs
```diff
@@ -0,0 +1,219 @@
...
+            return;
+
+        if (!TryComp(args.Used, out PowerCauteryComponent? cautery))
+            return;
+
+        if (!TryComp(args.Used, out BatteryComponent? battery))
+            return;
+
+        // get target bloodstream
+        TryComp<BloodstreamComponent>(target, out var bloodstream);
```
SnappingOpossum: This could be moved down and replace the null check.

Content.Shared/_Carpmosia/Healing/Systems/PowerCauterySystem.cs
```diff
@@ -0,0 +1,219 @@
...
+        if (!powered)
+        {   // starting the doafter w/o power is fine if we can fail it here
+            _popupSystem.PopupClient(Loc.GetString("power-cautery-no-power"), args.User, args.User);
+            return;
+        }
+
+        // Stem bleeding.
+        if (cautery.BloodlossModifier != 0 && bloodstream != null)
+        {
+            var isBleeding = bloodstream.BleedAmount > 0;
```
SnappingOpossum: This is unused.

Content.Shared/_Carpmosia/Healing/Systems/PowerCauterySystem.cs
```diff
@@ -0,0 +1,219 @@
...
+        // repeat if our target is still bleeding
+        args.Repeat = IsBleeding((args.Used.Value, cautery), target);
+
+        args.Handled = true;
+
+        // say we're finished if we're not repeating
+        if (!args.Repeat)
+        {
+            _popupSystem.PopupClient(Loc.GetString("medical-item-finished-using", ("item", args.Used)), target.Owner, args.User);
+            return;
```
SnappingOpossum: This will always en after this anyway, the return isn't needed.

Resources/Prototypes/_Carpmosia/Entities/Objects/Specific/Medical/healing.yml
```diff
@@ -5,3 +5,43 @@
...
+  - type: GenericVisualizer
+    visuals: # TODO: make it use all 5 steps of the battery indicator w/ some other visualizer
+      enum.BatteryVisuals.State: # because this one doesn't work
+        enum.PowerCellVisualLayers.Unshaded:
+          Full: {visible: true, state: battery-5}
+          Neither: {visible: true, state: battery-3}
+          Empty: {visible: false}
+  - type: Tag
+    tags:
+    - SurgeryTool # this isn't close to a surgery tool and won't be usable in surgeries but this needs to fit in medical belts somehow and I cba adjusting their whitelist
```
SnappingOpossum, line 34: Regular cautery is a surgery tool to it's close enough. Might fit parented to BaseToolSurgery.

SnappingOpossum (changes requested):

Content.Shared/_Carpmosia/Healing/Systems/PowerCauterySystem.cs
SnappingOpossum: I meant that that GetAndDrawPower itself should be able to be replaced with TryUseCharge.

CarpBot (changes requested):

PR #338, opened 2026-05-11
Remove nitrous price change and lower frozen price

SnappingOpossum:
## General information

Nitrous oxide has been reverted to a price per mole of 0.1, frezon had its price reduced to 1.5 per mole.

## Why / Balance

Requested by Celeste.

## Technical details

Basic YAML.

## Changelog

:cl:
- tweak: Nitrous oxide and frezon prices have been reduced.

### DIFF
--- a/Resources/Prototypes/Atmospherics/gases.yml
+++ b/Resources/Prototypes/Atmospherics/gases.yml
@@ -101,7 +101,7 @@
   molarMass: 44
   color: '#8F00FF'
   reagent: NitrousOxide
-  pricePerMole: 1 # Carpmosia-edit - More valuable gasses
+  pricePerMole: 0.1
 
 - type: gas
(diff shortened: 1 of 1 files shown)


CarpBot (changes requested):

wavesneglected:
Reasoning is that there should be no world where frezon is more expensive than tritium. I do agree that Wizards Den has their price too low however it should not be high enough to just defeat Cargo's gameplay loop. Hopefully this may in the far future encourage Atmos to do more than huff cash and post money spreads on the internet.

Atakku (approved):
mrrp miau

imatsoup (approved):
some may weep for spesoborg but i believe they will be just fine

CarpBot (approved):

Issue #326, opened 2026-05-07
Engine Looses need to be Less Awful


imatsoup:
There's been a lot of talk about making engine looses not totally round obliterating (in particular the Tesloose). I was talking with Celeste (among others) about what exactly a loose should and should not be and I thought it would be a good idea to leave an issue here on the subject. I do plan to try and come up with something after I'm done with my current project.

An engine loose should:

- Be a huge disaster for the station to cleanup.
- Have the potential to end the round if it isn't cleaned up well, the crew lacks the required personnel, or if it otherwise occurs on top of some other disaster
- Be clearly telegraphed (upstream already did some of the work for us there with the Emitter changes, but we could make a similar change to the PA to alert people when its been tampered with or activated)

An engine loose should *not*:
- Be a guaranteed end-of-round scenario (looking at you Tesloose)
- Do so little damage as to only be of concern to engineering
- Only be solvable with specialized equipment locked behind science (decels)

Reworking looses to meet these requirements would go a long way toward removing the need for excessive destruction rules, which is highly desired. Some ideas put forward:

- The Tesla should have a more reliable method in which it self-terminates. Celeste came up with a pretty interesting idea to have it absorb energy from power sources and then explode spectacularly after absorbing a certain amount, but that is a very serious rewrite

- The Tesla probably shouldn't just hunt players (as funny as the lightning ball being able to sense fear is)

- The Singulo should probably have a similar method of timing out once it does a certain amount of destruction. This conflicts with current behavior (eating stuff makes it last longer currently)

This should be considered a living issue that updates as talks continue and we work out what we want other requirements to be.

wavesneglected:
<img width="1706" height="378" alt="Image" src="https://github.com/user-attachments/assets/79a9c688-3563-458d-97c6-de1f5d338cde" />

Idea provided by Cybertabby, unsure how something would be implemented like this but I like the idea.

PR #341, opened 2026-05-12
PA radio alerts

imatsoup:
## General information

Adds radio alerts for when the PA is turned off and on, as well as if the PA strength is set to 4 (the one accessed by cutting a wire).
## Why / Balance

Part of the work for https://github.com/carpmosia/carpmosia/issues/326. Makes PA sabotage no longer a silent bit of sabotage someone can do to wreck the station one way or another (either through killing an engine or loosing it).
## Technical details

Added a method to the ParticleAcceleratorBoxSystem for sending radio alerts and added some loc strings to the component. (Most of this stuff is ripped from the Emitters)
## Media

image

## Breaking changes

None that I'm aware of.
## Changelog

:cl:
- add: The particle accelerator will now alert engineering if it is activated, deactivated, or if it's strength is set to 4.

### DIFF
--- a/Content.Server/ParticleAccelerator/Components/ParticleAcceleratorControlBoxComponent.cs
+++ b/Content.Server/ParticleAccelerator/Components/ParticleAcceleratorControlBoxComponent.cs
@@ -135,7 +135,7 @@ public sealed partial class ParticleAcceleratorControlBoxComponent : Component
     [ViewVariables]
     public bool CanBeEnabled = true;
 
-// Carpmosia-start - PA radio alerts
+    // Carpmosia-start - PA radio alerts
     /// <summary>
     /// The radio channel to broadcast on when something happens to this PA
     /// </summary>
@@ -165,6 +165,6 @@ public sealed partial class ParticleAcceleratorControlBoxComponent : Component
     /// </summary>
     [DataField]
     public LocId LocWarning = "pa-warning-broadcast";
-// Carpmosia-end - PA radio alerts
+    // Carpmosia-end - PA radio alerts
 
 }
--- a/Content.Server/ParticleAccelerator/Components/ParticleAcceleratorControlBoxComponent.cs
+++ b/Content.Server/ParticleAccelerator/Components/ParticleAcceleratorControlBoxComponent.cs
@@ -146,7 +146,7 @@ public sealed partial class ParticleAcceleratorControlBoxComponent : Component
     /// Whether a radio channel should be alerted if anything happens to this PA
     /// </summary>
     [DataField]
-    public bool AlertRadio = false;
+    public bool AlertRadio = true;
 
     /// <summary>
     /// Localized string to use when this particle accelerator is activated and AlertRadio is set to true
--- a/Content.Server/ParticleAccelerator/Components/ParticleAcceleratorControlBoxComponent.cs
+++ b/Content.Server/ParticleAccelerator/Components/ParticleAcceleratorControlBoxComponent.cs
@@ -1,5 +1,7 @@
 using Content.Server.ParticleAccelerator.Wires;
+using Content.Shared.Radio; // Carpmosia-edit - PA radio alerts
 using Content.Shared.Singularity.Components;
+using Robust.Shared.Prototypes; // Carpmosia-edit - PA radio alerts
 using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom;
 
 namespace Content.Server.ParticleAccelerator.Components;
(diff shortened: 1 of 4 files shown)


CarpBot (changes requested):

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
```diff
@@ -436,4 +444,15 @@ public static int GetPANumericalLevel(ParticleAcceleratorPowerState state)
...
+// Carpmosia-start - PA radio alerts
+        private void AlertRadio(ParticleAcceleratorControlBoxComponent comp, string locString)
+        {
+            if (!comp.AlertRadio)
+                return; // APEs do not need to scream over engineering radio, and an emitter that is off is probably not going to be alerting radios
+
+            var message = Loc.GetString(locString);
+            _radio.SendRadioMessage(comp.Owner, message, comp.RadioChannel, comp.Owner);
+        }
+// Carpmosia-end - PA radio alerts
```
SnappingOpossum: Indentation.
```suggestion
    // Carpmosia-start - PA radio alerts
    private void AlertRadio(ParticleAcceleratorControlBoxComponent comp, string locString)
    {
        if (!comp.AlertRadio)
            return; // APEs do not need to scream over engineering radio, and an emitter that is off is probably not going to be alerting radios

        var message = Loc.GetString(locString);
        _radio.SendRadioMessage(comp.Owner, message, comp.RadioChannel, comp.Owner);
    }
    // Carpmosia-end - PA radio alerts
```

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
```diff
@@ -436,4 +444,15 @@ public static int GetPANumericalLevel(ParticleAcceleratorPowerState state)
             _ => 0
         };
     }
+
+// Carpmosia-start - PA radio alerts
+        private void AlertRadio(ParticleAcceleratorControlBoxComponent comp, string locString)
```
SnappingOpossum: Pass in the entity as well instead of using component.Owner.
Also couldn't the string here be a LocId?

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
```diff
@@ -193,6 +199,8 @@ or ParticleAcceleratorPowerState.Level1
                         AudioParams.Default.WithVolume(-8f));
                     comp.EffectCooldown = _gameTiming.CurTime + comp.CooldownDuration;
                 }
+
+                AlertRadio(comp, comp.LocWarning); // Carpmosia-edit - PA radio alerts
```
SnappingOpossum: I feel like we shouldn't be tying this to the level at which it alerts admins.

Content.Server/ParticleAccelerator/Components/ParticleAcceleratorControlBoxComponent.cs
```diff
@@ -132,4 +134,37 @@ public sealed partial class ParticleAcceleratorControlBoxComponent : Component
...
+    /// The radio channel to broadcast on when something happens to this PA
+    /// </summary>
+    [DataField]
+    public ProtoId<RadioChannelPrototype> RadioChannel = "Engineering";
+
+    /// <summary>
+    /// Whether a radio channel should be alerted if anything happens to this PA
+    /// </summary>
+    [DataField]
+    public bool AlertRadio = false;
```
SnappingOpossum: Why not have this default to true?

SnappingOpossum (dismissed):

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
Atakku: idk thats how EmitterSystem.cs does it, isnt it basically the same type just aliased?

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
imatsoup: I mean, its the strength you can only access by snipping the wire so. We would just be checking the same thing again so it makes sense to have it there. Settings all the way up to 3 probably shouldn't cause an alert given that they are "intended"

Content.Server/ParticleAccelerator/Components/ParticleAcceleratorControlBoxComponent.cs
imatsoup: Dunno! I patterned this off of the emitters, which also does this. I presume its for a theoretical version of it that does Not alert the radio for some reason or another.

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
imatsoup: I forgot that doing it the way I was doing it was deprecated :melting_face:

Content.Server/ParticleAccelerator/Components/ParticleAcceleratorControlBoxComponent.cs
SnappingOpossum: I think that's on the emitters for the APE, since that counts as an emitter.

imatsoup: oh word i hadn't considered that, went ahead and changed it to default to true

Atakku (approved):
mrrp miau

CarpBot (approved):

PR #342, opened 2026-05-12
Quick fix to PA radio alerts

Atakku:
## General information

Small fix

## Why / Balance

Small fix

## Technical details

Made PA level alert regardless of admin log, just as @SnappingOpossum suggested


CarpBot (changes requested):

arenaconspiracy:
Why not just bolt this onto the original PR. Before it was merged I mean.

Atakku:
> Why not just bolt this onto the original PR. Before it was merged I mean.

im eepy

wavesneglected:
because we didn't have the conversation til i mentioned it in VC. It was a Rushed merge.

imatsoup:
I'm a bit confused. Why are we essentially retesting for the exact same thing as the statement above?

Atakku:
> I'm a bit confused. Why are we essentially retesting for the exact same thing as the statement above?

cvar could be changed

Atakku:
also this alerts if anything else in the  future flips the PA other than the player

PR #360, opened 2026-05-26
Cleanup from the upstream merge

imatsoup:
## General information

Cleans up some things with the upstream merge, particularly around bloodbound.
## Why / Balance

cat noises


SnappingOpossum:
Linter fails real, the damageContainer fields need to be on an Injurable component, the SalvageSpecialist needs the antag bad to come with the restricted antag part, and the solution field can be safely deleted.

More for later, but I'd also like to see the blood bound conversion use an AntagSpecifier.

imatsoup:
i will be fixing the linter fails in a moment had to step away

Resources/Prototypes/_Carpmosia/Entities/Objects/Consumable/Drinks/drinks_metamorphic.yml
```diff
@@ -8,13 +8,12 @@
...
-  - type: SolutionContainerManager
-    solutions:
-      drink:
-        maxVol: 30
-        reagents:
-        - ReagentId: Angostura
-          Quantity: 30
+  - type: Solution
+    solution:
+      maxVol: 30
```
SnappingOpossum: These shouldn't be specifying a maxVol.

SnappingOpossum (changes requested):

Resources/Prototypes/_Carpmosia/Entities/Objects/Consumable/Food/produce.yml
```diff
@@ -9,15 +9,14 @@
...
-      food:
-        maxVol: 14
-        reagents:
-        - ReagentId: Oil
-          Quantity: 5
-        - ReagentId: Phosphorus
-          Quantity: 5
+  - type: Solution
+    solution:
+      maxVol: 14
```
SnappingOpossum: This shouldn't either.

Resources/Prototypes/_Carpmosia/Entities/Objects/Fun/plushies.yml
```diff
@@ -50,15 +50,14 @@
...
-      food:
-        maxVol: 20
-        reagents:
-        - ReagentId: Fiber
-          Quantity: 10
-        - ReagentId: JuiceThatMakesYouScream
-          Quantity: 10
+  - type: Solution
+    solution:
+      maxVol: 20
```
SnappingOpossum: And these.

SnappingOpossum:
I think what's left is map fixes for rotated grilles

Atakku:
> I think what's left is map fixes for rotated grilles

i will do those in a sec

Atakku (approved):
mrrp miau

Resources/Prototypes/_Carpmosia/Entities/Objects/Fun/plushies.yml
```diff
@@ -139,17 +137,13 @@
...
-      food:
-        maxVol: 20
-        reagents:
-        - ReagentId: Fiber
-          Quantity: 10
-        - ReagentId: Blood
-          Quantity: 10
-      coloring:
-        maxVol: 30
+  - type: Solution
```
SnappingOpossum, line 140: I should also note this plushie's colouring feature won't work properly now, we should consider what we want to do with it going forward.

Atakku: huh why

SnappingOpossum: It doesn't have a `coloring` solution. It's not urgent but something to keep in mind.

Atakku: ah ye thats fixable tho

SnappingOpossum (approved):

PR #354, opened 2026-05-20
Brain removal tool

imatsoup:
## General information

Adds the 'experimental auto-surgeon' (a sprite edit on jaws of life) to the robotech vendor and protolathe surgery packages.

## Why / Balance

Lifts the major blocker on https://github.com/carpmosia/carpmosia/issues/340 by introducing a tool explicitly for brain extraction to be used with MMIs.
## Technical details

Adds a new bespoke component/system for the auto-surgeon. Currently it only looks for and removes the one kind of organ. Refactoring it to be more generic would be trivial but there is currently no point in removing other organs. Ditto for *inserting* organs into bodies- this actually would open up a new revival method that I'm not certain I would want in the game.
## Media

https://github.com/user-attachments/assets/5e20519b-704b-40fd-b1e8-40470e0b255a

image

image

image

image

## Breaking changes

Doesn't touch existing systems, outside of adding a new listing to the robotech and a new item for the static surgical lathe pack, both of which are in yaml. This is purposeful to make removing it all as close to seamless as possible when discomed finally comes out way. 

## Changelog

:cl:
- add: The RoboTech Deluxe vendor and exosuit fabricator now hold the Experimental Auto-Surgeon for the purposes of brain removal with minimal effort on the part of the roboticist.

PR #362, opened 2026-05-27
Remove damage-threshold based body destruction.

### LINKED ISSUE #340: Gibbing and ashing removal requirements
The removal of blunt-based gibbing and heat-based ashing is a desired feature on carpmos. We feel removing of this behavior:
- Encourages traitors and other antagonists to think outside the box when figuring out how to go about round-removing targets
- Removes thoughtless round-removal as a possibility as a step toward removing 2.9 rules
- Makes atmos-based sabotages and mass-casualty events less devastating (the amount of raw heat damage a plasma or tritfire can do is still a problem but its a step in the right direction)

Removing these behaviors is relatively trivial, however we need to address the following:
- Removing of brains in order to create MMIs for borgs is a process that needs to be made doable without weird work-arounds like the meatspike. This should be considered the only hard requirement for the removal of blunt-based gibbing. Any change solving the borging issue may be PR'd separately or in the same PR.
- - To accomplish this, the current desired change is to have an interaction with the bonesaw when used on a humanoid corpse that is buckled onto a surgical bed, either removing the brain specifically or gibbing after a long do-after.
- Additionally, some traitor equipment designed for corpse disposal needs changing. This list is not comprehensive:
- - Northstars probably need a price reduction and possibly a new gimmick (making corpses unrecognizable after doing enough damage perhaps)
- - Minibombs may need some adjustments
- - Incendiary grenades may need some adjustments

imatsoup:
## General information

Does what it says.
## Why / Balance

https://github.com/carpmosia/carpmosia/issues/340, removing these behaviors (given they are essentially placeholders for a more in-depth damage and medical system) is desired. A necessary step forward for removal of excessive destruction rules and allowing antags more freedom to do terrible things. 
## Test plan

throngle some urists
## Changelog

:cl:
- remove: Removed damage based gibbing and ashing. Other methods still work for now.

### DIFF
--- a/Resources/Prototypes/Body/Species/vox.yml
+++ b/Resources/Prototypes/Body/Species/vox.yml
@@ -187,28 +187,30 @@
       amount: 5
   - type: Destructible
     thresholds:
-    - trigger:
-        !type:DamageTypeTrigger
-        damageType: Blunt
-        damage: 400
-      behaviors:
-      - !type:GibBehavior { }
-    - trigger:
-        !type:DamageTypeTrigger
-        damageType: Heat
-        damage: 1500
-      behaviors:
-      - !type:SpawnEntitiesBehavior
-        spawnInContainer: true
-        spawn:
-          FoodMeatChickenFriedVox:
-            min: 3
-            max: 5
-      - !type:BurnBodyBehavior
-        popupMessage: bodyburn-vox-text-others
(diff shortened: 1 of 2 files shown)


imatsoup:
requires https://github.com/carpmosia/carpmosia/pull/354

Atakku (approved):
mrrp miau

PR #366, opened 2026-05-29
Cream Pies dont force item drops

imatsoup:
## General information

Banana Cream Pies no longer force you to drop your items. 
## Why / Balance

Part of the war on stun meta. The Knockdown is still pretty good, it just wont let you easily yoink someone's gun or something. True Clowns will see almost no difference in behavior.
## Test plan

Spawn Urist, equip them with an item in their hands, throw a pie at them.
## Changelog

:cl:
- tweak: Getting hit by a cream pie will no longer force you to drop your held items!

### DIFF
--- a/Content.Shared/Nutrition/EntitySystems/SharedCreamPieSystem.cs
+++ b/Content.Shared/Nutrition/EntitySystems/SharedCreamPieSystem.cs
@@ -122,7 +122,7 @@ public abstract partial class SharedCreamPieSystem : EntitySystem
 
         // TODO: Check if they even have a head that can be hit.
         SetCreamPied(creamPied.AsNullable(), true);
-        _stunSystem.TryUpdateParalyzeDuration(creamPied.Owner, creamPie.ParalyzeTime);
+        _stunSystem.TryKnockdown(creamPied.Owner, creamPie.ParalyzeTime, true, true, false); // Carpmosia-edit - Cream pies don't drop items
 
         // Throwing is not predicted, so the thrower is not equal to the client predicting the collision, so we cannot pass in a user.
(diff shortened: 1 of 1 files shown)


arenaconspiracy:
This reminded me that tazers should probably have a larger projectile hitbox, since tazers are essentially short range battery powered pies.
It could be argued that pies not disarming you actually has bigger combat utility for fact you're slower on the ground with filled hands, but it's probably better this way.

Atakku (approved):
mrrp miau

Content.Shared/Nutrition/EntitySystems/SharedCreamPieSystem.cs
```diff
@@ -122,7 +122,7 @@ private void OnCreamPiedHitBy(Entity<CreamPiedComponent> creamPied, ref ThrowHit
 
         // TODO: Check if they even have a head that can be hit.
         SetCreamPied(creamPied.AsNullable(), true);
-        _stunSystem.TryUpdateParalyzeDuration(creamPied.Owner, creamPie.ParalyzeTime);
+        _stunSystem.TryKnockdown(creamPied.Owner, creamPie.ParalyzeTime, true, true, false); // Carpmosia-edit - Cream pies don't drop items
```
SnappingOpossum: You should be able to just do this instead of setting the others as well.
```suggestion
        _stunSystem.TryKnockdown(creamPied.Owner, creamPie.ParalyzeTime, drop: false); // Carpmosia-edit - Cream pies don't drop items
```

SnappingOpossum (changes requested):

SnappingOpossum (approved):

PR #371, opened 2026-05-31
Change server description.

SnappingOpossum:
## General information

Removed the "We just launched" from the description.

## Why / Balance

We have not, in fact, just launched.

## Changelog

Change for server browser, not in game.


SnappingOpossum:
Forgot to draft it.

Atakku:
Pretty sure we had a better desc in the works

Atakku (approved):
mrrp miau

Issue #327, opened 2026-05-07
Centcomm needs reworking


imatsoup:
We changed eorg rules to ban it but haven't updated centcomm yet. Need to update centcomm to disincentivize eorging and bring things up to current standards. CarpComm should have the following changes:

- Removal of the current armory and replacing it with an Amnesty box that can serve as a dumping ground for contraband, sec gear etc
- Removal of borderline debug items like the Throngler
- Making CC equipment inaccessible to station command staff (the central office shouldn't have Command perms and instead should only open to CC staff)
- CC medbay needs a minor overhaul
- - The cloning facilities need to be removed
- - Carp has expanded cryochems and they could be represented in the cryo room more (maybe include necrosol or something if you want to be lazy)

wavesneglected:
After I finish the things I have eating me on my plate, I will discuss this further with my circle and perhaps find someone to tackle this.

PR #375, opened 2026-06-01
Centcomm tweaks

imatsoup:
## General information

Makes tweaks to central command to be more in line with what we want out of it.
## Why / Balance

We banned eorg but CC wasn't changed to support that, thus https://github.com/carpmosia/carpmosia/issues/327. There was also an opportunity to add some support for the cryomeds we have into the CC medbay. 
## Technical details

Changed map, saved to our own map namespace, changed relevent paths.
## Test plan

load any map, make sure cc also loads with it.
## Changelog

:cl:
- tweak: Centcomm has been modified to make EORG less likely


Resources/Maps/_Carpmosia/centcomm.yml
```diff
@@ -49046,120 +49046,6 @@ entities:
...
-  - uid: 7312
-    components:
-    - type: MetaData
-      name: Panic Switch
-    - type: Transform
-      pos: 1.5,2.5
-      parent: 1
-    - type: DeviceNetwork
-      configurators:
-      - invalid
```
SnappingOpossum: You only needed to delete these lines.

Atakku:
the diff seems to be
image
image

Atakku (approved):
mrrp miau

Atakku:
fuck it we ball

PR #378, opened 2026-06-03
Improve round restart vote logic

arenaconspiracy:
By GUTTING it like a fish.
## General information
Simplifies round restart votes by 1) making them callable at any time 2) removing abstinence 3) removing admin presence logic.
Also knocks Yes vote threshold down to 75% down from 85% because I was asked to.

## Why / Balance
Necessary for future changes.

Vote abstinence logic is opaque and meaningless when Nay and Abs are functionally identical.
Rounds becoming unplayable don't _necessarily_ mean a server wipe, and existing logic includes skipping the ghost check below a certain population, making it further redundant.
The CCVar preventing restarts with online admins is _not telegraphed_ in chat - only in adminlogs - and I didn't even know it _existed_ before looking at the C# (no wonder everyone complains about the restart votes never working), and the functionality was deemed undesirable.

## Technical details
Comments out undesirable logic in `VoteManager.DefaultVotes` and adds one entry to the production config preset.

## Test plan
1) Launch as many game clients as you can without killing your computer.
2) `sudo cvar votes.timerrestart 10`
3) Run a bunch of restart votes and look at them.

## Media
Imagine it or something, I'm not launching more than 3 game clients to take a screenshot.

## Breaking changes
Several CCVars are now obsolete and unused and I'm not sure what to do with them.

## Changelog
:cl:
- tweak: You can no longer abstain from round restart votes.
- tweak: You can now call round restart votes at any time.
- tweak: Did you know that restart votes silently fail if an admin is present? Well, not anymore they don't.

PR #379, opened 2026-06-03
Remove shuttle controls from comms consoles

arenaconspiracy:
## General information
Removes emergency shuttle control ( call / recall ) elements from the Communications Console UI.
Additionally removes the shuttle call Ion Law because it's completely inactionable now.

Requires #378 
(Not functionally, but it would help with the __EDGE CASES__ where the round should reasonably end before 90 minutes)

## Why / Balance
(assisted in writing by ringlespore & lady death's head)

1) **Rounds should not last two or three or more hours.**
You wanted to play for 3 to 5 hours straight? We have a thing for that, it's called playing multiple rounds in a row.
In this game's nature as a disaster simulator, you should _want_ to get out of dodge by at most 60 minutes into the round, and _have to_ by at most 90. If neither occur by 90 then _something with the game_ has gone amiss, and you DON'T wait another 90 minutes to see if anything "evac-worthy" happens, you Cycle Out so the next round hopefully works as intended.
There's no legitimate reason for a round to last more than 90 minutes; character interactions can pick up at the next round, Money-Making and Gas-Production and Some-Such should be done _in service to a real round_ and do not justify carrying on a dead one.
In case someone gets removed from the round they will be consigned to an Underwhelming observer experience, or an Underwhelming selection of ghostroles, delaying new presence in the round works in nobody's service.
That's all not to mention that playing a 5 hour round interrupted is Medically Inadvisable at the best of times.

2) **Early evacuation erodes gameplay quality.**
Haphazard use of shuttle calling ruins the element of Disaster Simulator. You are always one keystroke away from erasing the very possibility of an engaging round.
An early evacuation can rob antagonists of Their Shot at antagonism, completely deny NukeOps or Traitors or whathaveyou of Gameplay, that they may have been waiting for days or weeks to get an attempt at.
Removing early evacuations removes the incentive for Traitors to cause "evac-worthy" damage, being another step towards removing Excessive Destruction rules.
Evacuating early lets Security give up and wait for the shuttle instead of persisting and excising large threats.
Evacuating early lets Engineering give up and wait for the shuttle instead of pulling through and fixing severe hull damage or power failures.
Game went amiss and nobody can recover the station unreasonably early? We have a solution to that, it's called a restart vote - confer  #378

3) **Shuttle votes erode roleplay quality.**
The status quo among the Command playerbase at the moment are kneejerk (re)call votes via comms consoles - either to the automatic call or at the first sight of anything going wrong - in the realm of "Do we recall Y for recall N for stay on common radio". And nothing short of a mass ban wave or taking away the option can effectively abate it.
Shuttle votes of this kind are inherently LRP, not just how through they're typically done, but because they subvert the worldbuilding & intended gameplay in that Command has control and the crew has to listen to them.

## Technical details
Comments out UI elements and corresponding UI logic, does not touch system logic but that can be included if deemed necessary.
Comments out an ion law prototype.

## Test plan
1) Left click on a Communications console.
2) Rejoice.

## Media
Also included is the station AI intrinsic announcement UI, which couldn't be used for calling shuttles anyway.
tmp

## Breaking changes
N/A

## Changelog
:cl:
- remove: You can no longer manually Call or Recall the Emergency Shuttle. Rounds will last no more than 90 minutes, and less only under extreme duress.

### DIFF
--- a/Content.Client/Communications/UI/CommunicationsConsoleMenu.xaml.cs
+++ b/Content.Client/Communications/UI/CommunicationsConsoleMenu.xaml.cs
@@ -70,14 +70,14 @@ namespace Content.Client.Communications.UI
 
             AlertLevelButton.Disabled = !AlertLevelSelectable;
 
-            EmergencyShuttleButton.OnPressed += _ => OnEmergencyLevel?.Invoke();
-            EmergencyShuttleButton.Disabled = !CanCall;
+            // EmergencyShuttleButton.OnPressed += _ => OnEmergencyLevel?.Invoke(); // Carpmosia-edit - No recalls
+            // EmergencyShuttleButton.Disabled = !CanCall; // Carpmosia-edit - No recalls
         }
 
         protected override void FrameUpdate(FrameEventArgs args)
         {
             base.FrameUpdate(args);
-            UpdateCountdown();
+            // UpdateCountdown(); // Carpmosia-start - No recalls
         }
 
         // The current alert could make levels unselectable, so we need to ensure that the UI reacts properly.
@@ -117,6 +117,7 @@ namespace Content.Client.Communications.UI
             }
         }
 
+        /*  // Carpmosia-start - No recalls
         public void UpdateCountdown()
         {
             if (!CountdownStarted)
@@ -133,5 +134,6 @@ namespace Content.Client.Communications.UI
                 ("time", diff.ToString(@"hh\:mm\:ss", CultureInfo.CurrentCulture)));
             CountdownLabel.SetMessage(infoText);
         }
+        */ // Carpmosia-end - No recalls
     }
 }
--- a/Content.Client/Communications/UI/CommunicationsConsoleBoundUserInterface.cs
+++ b/Content.Client/Communications/UI/CommunicationsConsoleBoundUserInterface.cs
@@ -85,10 +85,10 @@ namespace Content.Client.Communications.UI
                 _menu.CurrentLevel = commsState.CurrentAlert;
                 _menu.CountdownEnd = commsState.ExpectedCountdownEnd;
 
-                _menu.UpdateCountdown();
+                // _menu.UpdateCountdown(); // Carpmosia-start - No recalls
                 _menu.UpdateAlertLevels(commsState.AlertLevels, _menu.CurrentLevel);
                 _menu.AlertLevelButton.Disabled = !_menu.AlertLevelSelectable;
-                _menu.EmergencyShuttleButton.Disabled = !_menu.CanCall;
+                // _menu.EmergencyShuttleButton.Disabled = !_menu.CanCall; // Carpmosia-start - No recalls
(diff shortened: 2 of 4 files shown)


wavesneglected:
There are two sides of me.

One side thinks this is awesome, as I hate evac votes and rounds pushing on longer than they should. I'm

the other side of me wonders about complications regarding shuttle calls in other events.

 I'll think about it, no dissaproval but I do think this should be discussed among the admin team prior to considering being merged as it also somewhat relates to the admin guidelines .

Thanks Arena.

arenaconspiracy:
Most of the gamemodes feature automatic shuttle calls in Side Major Victory conditions ending the round, once all [NukeOps/Xorgs/Infected/Headrevs] are out of commission, or if [the nuke blows up/Xorgs get a supermajority/Infected get a supermajority/all of Command is dead or converted]. It could be worth looking into extending the Major Victory calls to the standard 10 minutes, as it currently forces it down to 5 so you can't recall it and extend a round without antagonists (does that sound familiar?)
Game administrators can still manually call/recall the shuttle via the F7 admin panel, so if a custom event requires the round to end or continue it's still at the administrator's discretion.

arenaconspiracy:
It was brought up by some people that the Shuttle Call ion law needs to be removed in tandem with this change, I'll get on that soon.

arenaconspiracy:
Ok done.

Atakku (changes requested):
I do not like this idea. This completely guts all longshift possibilities, and just makes us another wiz/goob clone which strive for shorter quicker rounds. A lot of people enjoy occasional longshifts, and they can be incredibly fun. Also longshifts are more common on lower pop, which helps players learn the game, try out some new things, and generally have fun playing around with stuff.

Atakku:
> It was brought up by some people that the Shuttle Call ion law needs to be removed in tandem with this change, I'll get on that soon.

I do support the shuttle call ion law to be removed though by itself, it can be extremely annoying depending on the person's interpretation

Atakku:
perhaps instead, a much much simpler and better implementation would be to lock the call button until a minimum round length? this would enforce a minimum round length, stop rounds from early evacuating from ramping event sheduler, making sure rounds last at least a certain amount of time, while still allowing people to extend a round when wanted, or call evac slightly earlier (depending on difference between minimum and first autocall)

Content.Client/Communications/UI/CommunicationsConsoleMenu.xaml.cs
SnappingOpossum, line 27:
```suggestion
        // public event Action? OnEmergencyLevel; // Carpmosia-edit - No recalls
```

SnappingOpossum: Okay I was gonna apply that but I'm not gonna trust it looking like that.

Atakku:
closing as we are not interested in this change.

PR #384, opened 2026-06-09
Health analyzer bloodstream reagents

CMuncey:
### Incomplete, do not merge
- Add science research
- Make MedTek+ cartridge printable (or craftable in G menu using a printable upgrade chip)
- Make CMO PDA start with MedTek+ cartridge
- Add new version of health analyzer with HealthAnalyzerPlus component (if wanted)

## General information
Adds a new variant of the health analyzer intended to be unlockable with science research, and given to CMO at roundstart in their PDA. The new version shows the reagents in the patients bloodstream as shown below. Since this uses a new HealthAnalyzerPlus ui, the cryo pods are not affected by the change and currently have no in-game way to get the HealthAnalyzerPlus component.

## Why / Balance
Makes treatment a little easier, as you can tell when reagents are used up to know when it is safe to apply more or apply other reagents that would have mixed like brute medicines.

This has potential to trivialize med a little too much, but maybe it won't be that big of a deal. The main thing it removes is the need to count to 10 for most meds. It also lessens the need to communicate the medicines you're injecting but I'm not convinced that it will go away due to this.

## Technical details
- Moved some containers around in the health analyzer gui
- The HealthAnalyzerUiState object now expects two Solution objects passed into it's constructor, one for the blood type (the soltuion that makes up the patients blood), and the bloodstream (the solution of all reagents in the blood)
  - The blood type is needed so we can ignore the patient's blood reagent(s) when presenting the reagents present

## Test plan
Just used the health analyzer while injecting a bunch of drugs and dealing various types of damage. Tested the cryopod too but not as extensively. More testing suggestions would be appreciated, there may be edge cases I missed.

## Media
image

## Breaking changes
Added some namespaces (I think? They would be called includes in C, not sure if C# has a different name for that or if it means namespace like C++) to some files so that I could use some object definitions.

Changed class constructor for HealthAnalyzerUiState

## Changelog
:cl:
- add: Added section to health analyzer ui to display the reagents present in the patients bloodstream
- tweak: Moved damage indicator list in the health analyzer ui to beside the new reagent list (functionality was not changed)


Content.Client/HealthAnalyzer/UI/HealthAnalyzerControl.xaml.cs
```diff
@@ -139,9 +145,17 @@ public void Populate(HealthAnalyzerUiState state)
                 .OrderByDescending(damage => damage.Value)
                 .ToDictionary(x => x.Key, x => x.Value);
 
+        // Carpmosia-start - Health analyzer bloodstream reagents
```
SnappingOpossum: This can be moved down a little.

Content.Client/HealthAnalyzer/UI/HealthAnalyzerControl.xaml.cs
```diff
@@ -93,18 +95,22 @@ public void Populate(HealthAnalyzerUiState state)
...
             _entityManager.TryGetComponent<MobStateComponent>(target.Value, out var mobStateComponent)
                 ? GetStatus(mobStateComponent.CurrentState)
                 : Loc.GetString("health-analyzer-window-entity-unknown-text");
 
+        // Carpmosia-start - Health analyzer bloodstream reagents
         // Total Damage
 
-        DamageLabel.Text = _damageable.GetTotalDamage(target.Value).ToString();
+        // DamageLabel.Text = _damageable.GetTotalDamage(target.Value).ToString();
+        // Carpmosia-end - Health analyzer bloodstream reagents
```
SnappingOpossum: Just do this an an inline comment.

Content.Client/HealthAnalyzer/UI/HealthAnalyzerControl.xaml.cs
```diff
@@ -156,10 +170,20 @@ private static string GetStatus(MobState mobState)
...
 
     private void DrawDiagnosticGroups(
+        FixedPoint2 totalDamage, // Carpmosia-edit - Health analyzer bloodstream reagents
         Dictionary<ProtoId<DamageGroupPrototype>, FixedPoint2> groups,
         IReadOnlyDictionary<ProtoId<DamageTypePrototype>, FixedPoint2> damageDict)
     {
-        GroupsContainer.RemoveAllChildren();
+        DamageGroupsContainer.RemoveAllChildren();
+
+        // Carpmosia-start - Health analyzer bloodstream reagents
```
SnappingOpossum: This should be moved up.

Content.Client/HealthAnalyzer/UI/HealthAnalyzerControl.xaml.cs
```diff
@@ -180,7 +204,7 @@ private void DrawDiagnosticGroups(
 
             groupContainer.AddChild(CreateDiagnosticGroupTitle(groupTitleText, damageGroupId));
 
-            GroupsContainer.AddChild(groupContainer);
+            DamageGroupsContainer.AddChild(groupContainer);
```
SnappingOpossum: Needs a comment.

Content.Client/Medical/Cryogenics/CryoPodWindow.xaml.cs
```diff
@@ -88,9 +88,9 @@ public void Populate(CryoPodUserMessage msg)
         // Health analyzer
         var maybePatient = _entityManager.GetEntity(msg.Health.TargetEntity);
         var hasPatient = msg.Health.TargetEntity.HasValue;
-        var hasDamage = hasPatient && msg.HasDamage;
+        // var hasDamage = hasPatient && msg.HasDamage; # Carpmosia-edit - Health analyzer bloodstream reagents
```
SnappingOpossum: Same for the one below.
```suggestion
        // var hasDamage = hasPatient && msg.HasDamage; // Carpmosia-edit - Health analyzer bloodstream reagents
```

Content.Server/Medical/HealthAnalyzerSystem.cs
```diff
@@ -236,13 +237,16 @@ public HealthAnalyzerUiState GetHealthAnalyzerUiState(EntityUid? target)
         var bloodAmount = float.NaN;
         var bleeding = false;
         var unrevivable = false;
+        Solution? bloodType = null; // Carpmosia-edit - Health analyzer bloodstream reagents
+        Solution? bloodSolution = null; // Carpmosia-edit - Health analyzer bloodstream reagents
 
         if (TryComp<BloodstreamComponent>(entity, out var bloodstream) &&
             _solutionContainerSystem.ResolveSolution(entity, bloodstream.BloodSolutionName,
-                ref bloodstream.BloodSolution, out var bloodSolution))
+                ref bloodstream.BloodSolution, out bloodSolution))
```
SnappingOpossum: Comment.

Resources/Locale/en-US/medical/components/health-analyzer-component.ftl
SnappingOpossum, line 1: FTL also supports comments, you can comment things out with a # or have a # above for single line changes, multiline should use surround comment style and ##

SnappingOpossum (changes requested):
They're called using directives, adding them isn't a breaking change.

I personally don't think it's good to have this but for now here's some nitpicking on the comments.

Resources/Locale/en-US/medical/components/health-analyzer-component.ftl
SnappingOpossum: Also new stuff can be put in the _Carpmosia subdirectory instead of being put here.

CMuncey: Will it automatically grab two sources if I add a new ftl file, or would I need to go in and update something to let it know where the new one is?

CMuncey:
> I personally don't think it's good to have this

I'm on the fence myself. It's cool info to have, and IC seems like it would exist seems like something the health analyzer would be able to do, but I can definitely see the argument of it making med too trivial

Resources/Locale/en-US/medical/components/health-analyzer-component.ftl
SnappingOpossum: Like with prototypes, as long as the root is correct they can be anywhere inside it.

Content.Client/HealthAnalyzer/UI/HealthAnalyzerControl.xaml.cs
```diff
@@ -201,6 +225,84 @@ private void DrawDiagnosticGroups(
         }
     }
 
+    // Carpmosia-start - Health analyzer bloodstream reagents
+    private void DrawBloodstreamInfo( float bloodLevel, Solution? bloodType, Solution? bloodSolution )
+    {
+        BloodstreamReagentsContainer.RemoveAllChildren();
+
+        List<ReagentQuantity> bloodstream = new List<ReagentQuantity>();
```
SnappingOpossum: You can make this just `var bloodstream` or have it as `new()` to remove redundant type specification.

Content.Client/HealthAnalyzer/UI/HealthAnalyzerControl.xaml.cs
```diff
@@ -201,6 +225,84 @@ private void DrawDiagnosticGroups(
...
+        FixedPoint2 bloodlevel_amount = 0;
+        FixedPoint2 bloodstream_reagent_amount = 0;
```
SnappingOpossum: The names should use camel case (applies to variables below as well) and the amounts can be set to FixedPoint2.Zero instead.

Content.Client/HealthAnalyzer/UI/HealthAnalyzerControl.xaml.cs
```diff
@@ -201,6 +225,84 @@ private void DrawDiagnosticGroups(
...
+
+        if ( bloodType is not null || bloodSolution is not null )
```
SnappingOpossum: Do this instead of the null suppression.
```suggestion
        if ( bloodType is not null && bloodSolution is not null )
```

Content.Client/HealthAnalyzer/UI/HealthAnalyzerControl.xaml.cs
```diff
@@ -201,6 +225,84 @@ private void DrawDiagnosticGroups(
...
+            ( "amount", bloodlevel_amount.ToString() ),
+            ( "percent", bloodlevel_percent.ToString() )
+        );
+
+        BloodstreamLabel.Text = Loc.GetString(
+            "health-analyzer-window-entity-bloodstream-text",
+            ( "amount", bloodstream_reagent_amount.ToString() )
+        );
+
+        NoReagents.Visible = ( bloodstream.Count == 0 );
```
SnappingOpossum: Parentheses aren't needed here.

Content.Client/HealthAnalyzer/UI/HealthAnalyzerControl.xaml.cs
```diff
@@ -201,6 +225,84 @@ private void DrawDiagnosticGroups(
...
+
+        foreach ( var reagent in bloodstream )
```
SnappingOpossum: You can deconstruct this into it's parts here by replacing it with something like `var (reagentId, reagentQuantity)`

Content.Client/HealthAnalyzer/UI/HealthAnalyzerControl.xaml.cs
```diff
@@ -201,6 +225,84 @@ private void DrawDiagnosticGroups(
...
+        {
+            var localized_name = _prototypes.Index<ReagentPrototype>( reagent.Reagent.Prototype ).LocalizedName;
+            var reagent_color = _prototypes.Index<ReagentPrototype>( reagent.Reagent.Prototype ).SubstanceColor;
```
SnappingOpossum: Just index this once and then grab the stuff.

Content.Client/HealthAnalyzer/UI/HealthAnalyzerControl.xaml.cs
```diff
@@ -201,6 +225,84 @@ private void DrawDiagnosticGroups(
...
+            var reagent_quantity = reagent.Quantity;
+
+            // Build the string for the current reagent
+            // Uses a unicode block character to show the reactant color
+            var reagent_name_num_string = Loc.GetString(
+                "health-analyzer-window-bloodstream-reagent-text",
+                ( "reagentColor", reagent_color ),
+                ( "reagentName", localized_name ),
+                ( "amount", reagent_quantity )
+            );
```
SnappingOpossum: Since you're only using this once you can just do it in its usecase instead.

SnappingOpossum (changes requested):

Content.Client/HealthAnalyzer/UI/HealthAnalyzerControl.xaml.cs
CMuncey: Oops, guess that was left over from an "is null" check, forgot to swap the op when negating

Atakku:
> > I personally don't think it's good to have this
> 
> I'm on the fence myself. It's cool info to have, and IC seems like it would exist seems like something the health analyzer would be able to do, but I can definitely see the argument of it making med too trivial

Maybe this could be implemented as advanced medical scanner, and maybe given to CMO or/and available with tier 2/3 science research

GlupiyBrush (changes requested):
1. All new sprites should be under the /Texture/_Carpmosia directory
2. Copyright in meta.json should reflect the fact that you made changes to the sprite

imatsoup:
honestly, after thinking for a while i'm kinda into it, but i wanna hear what your thoughts are in specific @SnappingOpossum 

also @CMuncey  address brub's comments so we can do more review

SnappingOpossum:
> honestly, after thinking for a while i'm kinda into it, but i wanna hear what your thoughts are in specific @SnappingOpossum

Honestly after thinking about it more I think I'd be fine with this even for regular medical staff, it's practically necessary to have multiple people somewhat effectively work on the same patient, even if it makes some things easier.

Atakku:
> Honestly after thinking about it more I think I'd be fine with this even for regular medical staff, it's practically necessary to have multiple people somewhat effectively work on the same patient, even if it makes some things easier.

Reverted to pre-medplus state, as I too thought about it week-two ago too, but didn't have time to dive into this PR

Atakku:
image

:fearful:

CarpBot (changes requested):

SnappingOpossum (approved):

PR #359, opened 2026-05-25
Block stripping multiple items at a time

arenaconspiracy:
## General information
Cancels humanoid stripping doafters if you try to begin a new doafter before finishing another.

## Why / Balance
Stripping a dead body under pressure should be a risk you take when you absolutely need to take something off of it.

As it stands, some passenger can come out of thin air and take off Every Single Item at a moment's notice off of a combatant the second they go critical.
Some passenger can walk into triage and leave someone bare within seconds before anyone can stop them.

## Technical details
Simple amendment of doafter arguments.

## Media
https://github.com/user-attachments/assets/77321e3f-ea67-4c61-9c07-7031eccba26a

## Breaking changes
N/A

## Changelog
:cl:
- tweak: You can no longer strip multiple items at a time off of people.

PR #388, opened 2026-06-11
Address some pain points with stripping items

imatsoup:
## General information

Follow up to https://github.com/carpmosia/carpmosia/pull/359.
## Why / Balance

Some early feedback has indicated that one of the pain points caused by our changes to stripping is security having to spend much more time brigging people. This change to stripping will allow you to strip people fully whilst they are in restraints without cancelling the doafter. 

Another less-intentional result of the change from 359 is that thief has received a nerf. I have decided to compensate them slightly by making them faster to strip people when pickpocketing. I want to make a further change to it to allow pickpocketing to be even faster as part of a slight rewrite to the stripping system, but that can come later.
## Technical details

Added a check to see if the target is restrained and dictate behavior of the doafter based on that.
## Test plan

Load into server, spawn urist, equip them fully, strip them with restraints and without.
## Changelog

:cl:
- tweak: Someone in restraints can have multiple items stripped off their person at once again.
- tweak: Thieves can now strip people slightly faster when pickpocketing

### DIFF
--- a/Content.Shared/Strip/SharedStrippableSystem.cs
+++ b/Content.Shared/Strip/SharedStrippableSystem.cs
@@ -703,12 +703,7 @@ public abstract partial class SharedStrippableSystem : EntitySystem
     // Carpmosia-start - Stripping changes
     public bool IsRestrained(EntityUid target)
     {
-        if (TryComp<CuffableComponent>(target, out var cuffed) && cuffed.CuffedHandCount > 0)
-        {
-            return true;
-        }
-
-        return false;
+        return TryComp<CuffableComponent>(target, out var cuffed) && cuffed.CuffedHandCount > 0;
     }
     // Carpmosia-end - Stripping changes
 }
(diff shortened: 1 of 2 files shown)


imatsoup:
hashtag kill all heisentests

Content.Shared/Strip/SharedStrippableSystem.cs
```diff
@@ -540,7 +540,7 @@ private void StartStripRemoveHand(
             BreakOnMove = true,
             NeedHand = true,
             BreakOnHandChange = false, // Allow simultaneously removing multiple items.
-            DuplicateCondition = DuplicateConditions.None // Carpmosia-edit - Slower stripping
+            DuplicateCondition = IsRestrained(target) ? DuplicateConditions.SameTool : DuplicateConditions.None  // Carpmosia-edit - Stripping changes
```
SnappingOpossum:
```suggestion
            DuplicateCondition = IsRestrained(target) ? DuplicateConditions.SameTool : DuplicateConditions.None // Carpmosia-edit - Stripping changes
```

SnappingOpossum (dismissed):
Cuffing a dead body to strip it faster

Atakku (approved):
mrrp miau

PR #393, opened 2026-06-12
Cleaned up map prototypes & Rotated Sparks & Lampocteis

Atakku:
## General information

Cleaned up map prototypes and added Sparks and Lampocteis into the map pool

## Why / Balance

Both maps were tested for quite some time now, and looking ready to be in rotation!

## Technical details

- [x] Rotated Sparks
- [x] Rotated Lampocteis
- [x] Cleaned up all map prototypes

## Test plan

No

## Media

No

## Breaking changes

No

## Changelog

:cl:
- tweak: Sparks is now in rotation!
- tweak: Lampocteis is now in rotation!


imatsoup (approved):
seems fine, though a bit unfortunate to rotate in new maps that probably wont get played until things die down 😢

Atakku:
> seems fine, though a bit unfortunate to rotate in new maps that probably wont get played until things die down 😢

monday is coming, i think we'd have at least a round

although its holiday season........

PR #394, opened 2026-06-12
Less annoying bureaucratic error

Atakku:
## General information

Bureaucratic has a fun little thing with 25% to nuke all roles but ONE (and passenger), while making that one infinite.
Adjusted it to instead make MORE roles infinite, while still nuking the rest. This should hopefully reduce the amount of job lockouts for late joiners, while keeping the absolute absurdity and fun of the rare event.

## Why / Balance

Make latejoin life better

## Technical details

- [x] Changed some Bureaucratic event logic 

## Test plan

`addgamerule BureaucraticError` several times until most jobs are locked out

## Media

image
image

## Breaking changes

None

## Changelog

:cl:
- tweak: BureaucraticError's extreme case of locking out every job has been tweaked to leave more (20-30%) of jobs unlocked


imatsoup (approved):
shore

PR #409, opened 2026-06-18
Cherry-pick Disposals refactor (#40540)

imatsoup:
## General information

Cherry-pick of https://github.com/space-wizards/space-station-14/pull/40540
## Why / Balance

Needed for the vent crawling port. Would usually wait for stable release, but there's no good way of knowing when that will be and I wanna work on vent crawling now.
## Breaking changes

i sure hope it doesnt


SnappingOpossum:
Might be better to try to grab everything from before things diverged instead, and since PRs have to be squashed I think that's best left to Atakku.

PR #411, opened 2026-06-20
Refactor mappools

Atakku:
## General information

Refactored map pools, used the CarpmosiaMapPool for carp-only maps, and added a LegacyMapPool for legacy maps

## Why / Balance

Cleanup, and utility

## Test plan

Change the cvar ingame and call a map vote to see that the new map pool is active

## Media

None

## Breaking changes

Carpmosia's default map pool has been renamed from CarpmosiaMapPool to DefaultMapPool, while the existing upstream DefaultMapPool has been renamed to UpstreamMapPool. Configs have been updated (map pool override removed) to accommodate this change.

## Changelog

No player facing changes


imatsoup (approved):
mapsz

PR #414, opened 2026-06-21
Fix dispos exploding

Atakku:
## General information

Changed minimum bounce threshold from 20 to 200 for disposals travel.

## Why / Balance

Upstream dispos rework has set a way too low margin of 20 turns before ripping the person/object out of dispos (and therefore breaking the loop). A lot of maps have way above 20, in the ballpack of 100 to actually complete a disposals loop, and this just keeps breaking dispos every time a trash is thrown inside.

## Test plan

Load up fland or smth and throw yourself in the trash. If you end up in disposals - good

## Media

No

## Breaking changes

No

## Changelog

:cl:
- fix: Fixed disposals unanchoring all the time


imatsoup (approved):
given this is a bugfix i say we get this in pronto

imatsoup:
well practically a bugfix

PR #424, opened 2026-06-26
Killed "Call shuttle" ion law

Atakku:
## General information

Removed the "call shuttle" ion law

## Why / Balance

Leads to generally annoying cases of borgs spamming evac shuttle very early on, and leads to their inevitable destruction.

## Technical details

- [x] Commented out the law.

## Test plan

if you really want you can ion storm yourself 5 billion times but this literally just comments it out so its not necessary

## Media

No

## Breaking changes

No

## Changelog

:cl:
- remove: The "THE SHUTTLE MUST BE CALLED BECAUSE OF (insert reason) ON THE STATION" ion law has been removed.

### DIFF
--- a/Resources/Prototypes/IonLaws/IonLaw.yml
+++ b/Resources/Prototypes/IonLaws/IonLaw.yml
@@ -5,11 +5,13 @@
     food: 0.5
     RandomManifestFill: 0.15
 
-- type: ionLaw
-  id: CallShuttle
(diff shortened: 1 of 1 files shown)


imatsoup (approved):
i racked my brain for a reason to keep this ion law and couldn't think of a good one. it's not even that funny half the time.

PR #434, opened 2026-06-27
Made all basic towels printable

Atakku:
## General information

Added all basic towel recipes to the lathes

## Why / Balance

They should be printable

## Technical details

- [x] Added recipes for towels and added them to the lathes

## Test plan

Check hop's uniform printer lathe in round

## Media

image

## Breaking changes

None

## Changelog

:cl:
- add: Most of the towels can now be printed at the uniform printer


imatsoup (approved):
sounds fine to me

PR #445, opened 2026-06-30
Converted rules to a one big read with collapsed sections

Atakku:
## General information

Unwrapped all of the rules and their extended clarifications into the main rule file.
Implemented collapsing sections for guidebook files, and wrapped all the clarifications in them.

## Why / Balance

I have personally witnessed how horrid the "click to read more" experience is, as it keeps jumping back to the top of the rules every time you press back and just demotivates people from reading the rules in their fullest entirely. So I changed to be them into one continuous read instead.

## Technical details

- [x] Reformatted rules into one big file.
- [x] Implemented folding, to hide clarifications by default.

## Test plan

Just start the server and click rules

## Media

https://github.com/user-attachments/assets/ac1d2fc8-5edd-4a76-a69e-e2e4a3c7951d

## Breaking changes

No

## Changelog

:cl:
- tweak: Rules have been reformatted for an easier continuous reading experience


Atakku:
Several current issues with our rules as well:
- AFAIK we do not require speaking English anywhere
- Metashield rules should be improved
- Dehumanizing should be added Core rules 2

imatsoup (approved):
is pretty neat

Content.Client/_Carpmosia/Guidebook/FakeLinkTag.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        var label = new Label();
+        label.Text = text;
+
+        label.MouseFilter = Control.MouseFilterMode.Pass;
+        label.FontColorOverride = LinkColor;
+        label.DefaultCursorShape = Control.CursorShape.Hand;
+
+        label.OnMouseEntered += _ => label.FontColorOverride = Color.LightSkyBlue;
+        label.OnMouseExited += _ => label.FontColorOverride = Color.CornflowerBlue;
```
SnappingOpossum, line 34: Could you make this take arbitrary colours and default to these?

SnappingOpossum (commented):
Current collapsible stuff seems to be fine so I don't think this will cause any issues itself.

Content.Client/_Carpmosia/Guidebook/FakeLinkTag.cs
Atakku: sure but in a followup pr, i literally copied this from the link tag 1:1

PR #448, opened 2026-06-30
Fix thief chance oopsie

Atakku:
## General information

Fixed accidental 95 chance instead of 75

## Why / Balance

was intending to change to 75 but forgor to update the 3 other places its used...
bugfix

## Technical details

no

## Test plan

no

## Media

no

## Breaking changes

no

## Changelog

no


Atakku:
johnselfmerg

PR #450, opened 2026-07-01
Readd aliases and extended, remove useless gamerule

SnappingOpossum:
## General information

Readded the game preset aliases, uncommented extended, and replaced the potluck gamerule with a `minPlayers: 40` on the preset.

## Why / Balance

Extended is useful for events since it doesn't run any major gamerules, but does have minor events happens that help things feel more normal than a greenshift. Additionally, it is used as a fallback game preset along with traitor, though I'm not sure how to properly test that so I can't comment on how much of an effect that has.

GamePresets support minPlayers, no need to use a gamerule for it.

## Breaking changes

Removed the `Potluck` gamerule and the file it was in.


Atakku (approved):
mrrp miau

PR #451, opened 2026-07-01
Added player_status_count metric

Atakku:
## General information

Added a player_status_count Prometheus metrics to count how many readied up players we have!

## Why / Balance

Metrics are good. Especially ready up metrics. Very good. Very useful.

## Technical details

- [x] Added `player_status_count` metric to count players in lobby, readied up, and in game

## Test plan

set `> cvar metrics.enabled True` on your test server and then check http://localhost:44880/metrics

## Media

No

## Breaking changes

No

## Changelog

No player facing changes

### DIFF
--- a/Content.Server/GameTicking/GameTicker.cs
+++ b/Content.Server/GameTicking/GameTicker.cs
@@ -97,6 +97,9 @@ namespace Content.Server.GameTicking
                 "Overflow role does not have the correct name!");
             InitializeGameRules();
             InitializeReplays();
+
+            InitializeMetrics(); // Carpmosia-edit - Extra prometheus metrics
+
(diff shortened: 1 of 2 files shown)


SnappingOpossum (dismissed):
Probably best to have these only run on release builds.

Atakku:
> Probably best to have these only run on release builds.

????

SnappingOpossum:
> ????

Just slap on like a `[Conditional("RELEASE")]` to the InitializeMertics and that should make it only run in release builds, there's not often a reason for running these outside those contexts and when you do need to you can just comment out that part to allow it anyway.

Atakku:
image

PR #453, opened 2026-07-01
Added names to map previews

Atakku:
## General information

Added names to all map preview entities

## Why / Balance

so they can look properly when ran through protovote

## Technical details

yaml

## Test plan

just protovote them

## Media

No

## Breaking changes

No

## Changelog

No player facing changes


Atakku:
jonh selfmerg

PR #457, opened 2026-07-03
Adds Nukeops Arrival Announcement

imatsoup:
## General information

Adds a special announcement when the nukie's shuttle FTLs into the sector on nukeops.
## Why / Balance

After removal of the cargo weapon crates, its become a commonly executed strategy to try to seize control of the armory before the crew has a chance to properly respond, usually as early into the round as possible. If successful, this would leave the operative team in a very strong position to win with very little resistance. While we do want to reward the execution of a plan, and there does come a point that the crew should probably just lose, I feel like most nukie rounds are more entertaining when both sides of the conflict get to participate fully. Seizing armory before the crew can arm is a strategy intended to very specifically stop the crew from being able to participate, and tends to be a strategy that plays out in pretty boring ways. So, as a consequence, the crew are now alerted when operative's shuttle first FTLs to the system. One unfortunate casualty of this change is stealth-ops, but I believe between this change and bringing the full armory crate category back, I would rather have this change.
## Test plan

Spawn nukeops, either by forcing the preset or addgameruling nukeops. Go to shuttle. FTL to station. The announcement plays and the station is now in red alert.
## Changelog

:cl:
- add: For nukeops, there is now an announcement that plays when the nukie shuttle first FTLs into the sector! This does not apply to Loneops.

### DIFF
--- a/Content.Server/GameTicking/Rules/Components/NukeopsRuleComponent.cs
+++ b/Content.Server/GameTicking/Rules/Components/NukeopsRuleComponent.cs
@@ -106,7 +106,7 @@ public sealed partial class NukeopsRuleComponent : Component
     /// Whether the arrival alert has played.
     /// </summary>
     [DataField]
-    public bool ArrivalAnnounced = false;
+    public bool ArrivalAnnounced;
 
     /// <summary>
     /// Alert levels that are 'below' red, used for changing it to red on FTL end
--- a/Content.Server/GameTicking/Rules/Components/NukeopsRuleComponent.cs
+++ b/Content.Server/GameTicking/Rules/Components/NukeopsRuleComponent.cs
@@ -100,6 +100,11 @@ public sealed partial class NukeopsRuleComponent : Component
     /// </summary>
     [DataField]
     public SoundSpecifier GreetSoundNotification = new SoundPathSpecifier("/Audio/Ambience/Antag/nukeops_start.ogg");
+
+    // Carpmosia-start - Nukeops Arrival Message
+    [DataField]
+    public bool ArrivalAnnounced = false;
+    // Carpmosia-end - Nukeops Arrival Message
 }
 
 public enum WinType : byte
--- a/Content.Server/GameTicking/Rules/Components/NukeopsRuleComponent.cs
+++ b/Content.Server/GameTicking/Rules/Components/NukeopsRuleComponent.cs
@@ -102,8 +102,17 @@ public sealed partial class NukeopsRuleComponent : Component
     public SoundSpecifier GreetSoundNotification = new SoundPathSpecifier("/Audio/Ambience/Antag/nukeops_start.ogg");
 
     // Carpmosia-start - Nukeops Arrival Message
+    /// <summary>
+    /// Whether the arrival alert has played.
+    /// </summary>
     [DataField]
     public bool ArrivalAnnounced = false;
+
+    /// <summary>
+    /// Alert levels that are 'below' red, used for changing it to red on FTL end
+    /// </summary>
+    [DataField]
+    public HashSet<string> Alerts = new HashSet<string>() { "green", "blue", "yellow", "violet" };
     // Carpmosia-end - Nukeops Arrival Message
 }
 
--- a/Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
@@ -45,7 +45,6 @@ using Robust.Shared.Random;
 using Robust.Shared.Utility;
 using System.Data;
 using System.Linq;
-using System.Reflection.Emit; // Carpmosia-edit - Nukeops Arrival Message
 using System.Text;
 
 namespace Content.Server.GameTicking.Rules;
--- a/Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
@@ -452,13 +452,29 @@ public sealed partial class NukeopsRuleSystem : GameRuleSystem<NukeopsRuleCompon
                 continue;
             if (!TryGetRandomStation(out var chosenStation))
                 return;
+            // Don't announce multiple times
+            if (nukeops.ArrivalAnnounced == true )
+                continue;
 
+            MapId? targetStationMap = null;
+            if (nukeops.TargetStation != null && TryComp(nukeops.TargetStation, out StationDataComponent? data))
+            {
+                var grid = data.Grids.FirstOrNull();
+                targetStationMap = grid != null
+                    ? Transform(grid.Value).MapID
+                    : null;
+            }
+            // Don't announce if we aren't going to the right place
+            if (targetStationMap == null || targetStationMap != Transform(ev.MapUid).MapID)
+                continue;
+            // Only set the alert level if its not already red
             if (_alertLevelSystem.GetLevel(chosenStation.Value) != "red")
-                _alertLevelSystem.SetLevel(chosenStation.Value, "red", true, true, true);
+                _alertLevelSystem.SetLevel(chosenStation.Value, "red", false, false, true);
 
             var msg = Loc.GetString("nukeops-shuttle-warning");
             _chat.DispatchGlobalAnnouncement(msg, playSound: false, colorOverride: Color.Red);
             _audio.PlayGlobal("/Audio/Misc/notice1.ogg", Filter.Broadcast(), true);
+            nukeops.ArrivalAnnounced = true;
         }
     }
     // Carpmosia-end - Nukeops Arrival Message
--- a/Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
@@ -35,6 +35,7 @@ using Content.Shared.Store.Components;
 using Content.Shared.Tag;
 using Content.Shared.Zombies;
 using Robust.Server.Player;
+using Robust.Shared.Audio;
 using Robust.Shared.Audio.Systems; // Carpmosia-edit - Nukeops Arrival Message
 using Robust.Shared.Containers;
 using Robust.Shared.Map;
--- a/Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
@@ -467,8 +467,10 @@ public sealed partial class NukeopsRuleSystem : GameRuleSystem<NukeopsRuleCompon
             // Don't announce if we aren't going to the right place
             if (targetStationMap == null || targetStationMap != Transform(ev.MapUid).MapID)
                 continue;
-            // Only set the alert level if its not already red
-            if (_alertLevelSystem.GetLevel(chosenStation.Value) != "red")
+                
+            // Only set the alert level if its not already red or higher
+            var alert = _alertLevelSystem.GetLevel(chosenStation.Value);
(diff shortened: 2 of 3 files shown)


Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
```diff
@@ -434,6 +443,27 @@ private void OnShuttleCallAttempt(ref CommunicationConsoleCallShuttleAttemptEven
...
+    {
+        var query = QueryActiveRules();
+        while (query.MoveNext(out var uid, out _, out var nukeops, out _))
+        {
+            if (ev.Entity != GetShuttle((uid, nukeops)))
+                continue;
+            if (!TryGetRandomStation(out var chosenStation))
+                return;
+            if (_alertLevelSystem.GetLevel(chosenStation.Value) == "red")
+                return;
```
Atakku: so.... if alert level was red, they never get the announcement?

imatsoup: should be fixed now

Resources/Locale/en-US/_Carpmosia/nukeops/shuttle.ftl
```diff
@@ -0,0 +1 @@
+nukeops-shuttle-warning = Warning! An FTL signature belonging to a Syndicate vessel was detected in your sector! Consider all personnel on board to be armed and dangerous.
```
Atakku:
```suggestion
nukeops-shuttle-warning = Warning! A Syndicate FTL signature has been detected in your sector! Consider all personnel on board to be armed and dangerous.
```

Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
```diff
@@ -33,21 +35,26 @@
...
 using Robust.Shared.Containers;
 using Robust.Shared.Map;
 using Robust.Shared.Network;
+using Robust.Shared.Player; // Carpmosia-edit - Nukeops Arrival Message
 using Robust.Shared.Prototypes;
 using Robust.Shared.Random;
 using Robust.Shared.Utility;
 using System.Data;
 using System.Linq;
+using System.Reflection.Emit; // Carpmosia-edit - Nukeops Arrival Message
```
Atakku: unused?
```suggestion
```

Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
```diff
@@ -434,6 +443,27 @@ private void OnShuttleCallAttempt(ref CommunicationConsoleCallShuttleAttemptEven
...
+            if (_alertLevelSystem.GetLevel(chosenStation.Value) != "red")
+                _alertLevelSystem.SetLevel(chosenStation.Value, "red", true, true, true);
+
+            var msg = Loc.GetString("nukeops-shuttle-warning");
+            _chat.DispatchGlobalAnnouncement(msg, playSound: false, colorOverride: Color.Red);
+            _audio.PlayGlobal("/Audio/Misc/notice1.ogg", Filter.Broadcast(), true);
+        }
+    }
+    // Carpmosia-end - Nukeops Arrival Message
+
```
Atakku, line 484: this should probably:
- trigger only once so nukies cant ftl back and forth to keep alerting it
- futureproof, where it announces this to the specific station nukies have FTL'd to (get all stations on map nukies FTL'd to)
but we can do that later when i implement multi stations ig so whateverrrrr

Atakku: Ok, priority major issue, this counts ANY ftl, EVEN ON THE SAME MAP right now. So if nukies HAVE to ftl multiple times to get to the station, people will get multiple alerts, which is not good!

Image

Atakku (changes requested):

Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
imatsoup: weird, i did test for that and on my local it didn't announce multiple times for some reason hrm

Atakku: > weird, i did test for that and on my local it didn't announce multiple times for some reason hrm

you probably tested that before
image
that

imatsoup: oh man i'm good at this

Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
SnappingOpossum: Probably should avoid changing it for any alert above red as well, don't wanna go from something like gamma or delta to this.

Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
```diff
@@ -434,6 +442,45 @@ private void OnShuttleCallAttempt(ref CommunicationConsoleCallShuttleAttemptEven
...
+                    ? Transform(grid.Value).MapID
+                    : null;
+            }
+            // Don't announce if we aren't going to the right place
+            if (targetStationMap == null || targetStationMap != Transform(ev.MapUid).MapID)
+                continue;
+                
+            // Only set the alert level if its not already red or higher
+            var alert = _alertLevelSystem.GetLevel(chosenStation.Value);
+            if (alert != "red" || alert != "gamma" || alert != "delta" || alert != "epsilon")
```
SnappingOpossum: Just do something like making a hashset of these and doing a .Contains check instead of 4 equivalence checks. Honestly might be good to consider futzing with the alert levels themselves as well, those seem to be fairly old and may not be entirely up to par, but that's for another time.

imatsoup: i humbly request re-review

Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
```diff
@@ -434,6 +442,45 @@ private void OnShuttleCallAttempt(ref CommunicationConsoleCallShuttleAttemptEven
...
+    {
+        var query = QueryActiveRules();
+        while (query.MoveNext(out var uid, out _, out var nukeops, out _))
+        {
+            if (ev.Entity != GetShuttle((uid, nukeops)))
+                continue;
+            if (!TryGetRandomStation(out var chosenStation))
+                return;
+            // Don't announce multiple times
+            if (nukeops.ArrivalAnnounced == true )
```
SnappingOpossum:
```suggestion
            if (nukeops.ArrivalAnnounced)
```

Content.Server/GameTicking/Rules/Components/NukeopsRuleComponent.cs
```diff
@@ -100,6 +100,20 @@ public sealed partial class NukeopsRuleComponent : Component
...
+    /// Whether the arrival alert has played.
+    /// </summary>
+    [DataField]
+    public bool ArrivalAnnounced = false;
+
+    /// <summary>
+    /// Alert levels that are 'below' red, used for changing it to red on FTL end
+    /// </summary>
+    [DataField]
+    public HashSet<string> Alerts = new HashSet<string>() { "green", "blue", "yellow", "violet" };
```
SnappingOpossum, line 115: Redundant type specification.
```suggestion
    public HashSet<string> Alerts = new() { "green", "blue", "yellow", "violet" };
```

Content.Server/GameTicking/Rules/Components/NukeopsRuleComponent.cs
```diff
@@ -100,6 +100,20 @@ public sealed partial class NukeopsRuleComponent : Component
     /// </summary>
     [DataField]
     public SoundSpecifier GreetSoundNotification = new SoundPathSpecifier("/Audio/Ambience/Antag/nukeops_start.ogg");
+
+    // Carpmosia-start - Nukeops Arrival Message
+    /// <summary>
+    /// Whether the arrival alert has played.
+    /// </summary>
+    [DataField]
+    public bool ArrivalAnnounced = false;
```
SnappingOpossum: This is the default value.
```suggestion
    public bool ArrivalAnnounced;
```

Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
```diff
@@ -434,6 +442,45 @@ private void OnShuttleCallAttempt(ref CommunicationConsoleCallShuttleAttemptEven
...
+                continue;
+
+            // Only set the alert level if its not already red or higher
+            var alert = _alertLevelSystem.GetLevel(chosenStation.Value);
+            if (nukeops.Alerts.Contains(alert))
+                _alertLevelSystem.SetLevel(chosenStation.Value, "red", false, false, true);
+
+            var msg = Loc.GetString("nukeops-shuttle-warning");
+            _chat.DispatchGlobalAnnouncement(msg, playSound: false, colorOverride: Color.Red);
+            _audio.PlayGlobal("/Audio/Misc/notice1.ogg", Filter.Broadcast(), true);
```
SnappingOpossum: Obsolete, use a SoundSpecifer or ResolvedSoundSpecifier. Something like this should work.
```suggestion
            _audio.PlayGlobal(new SoundPathSpecifier("/Audio/Misc/notice1.ogg"), Filter.Broadcast(), true);
```

Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
```diff
@@ -434,6 +442,45 @@ private void OnShuttleCallAttempt(ref CommunicationConsoleCallShuttleAttemptEven
...
+    // Carpmosia-start - Nukeops Arrival Message
+    private void OnShuttleFTLCompleted(ref FTLCompletedEvent ev)
+    {
+        var query = QueryActiveRules();
+        while (query.MoveNext(out var uid, out _, out var nukeops, out _))
+        {
+            if (ev.Entity != GetShuttle((uid, nukeops)))
+                continue;
+            if (!TryGetRandomStation(out var chosenStation))
+                return;
```
SnappingOpossum, line 455: Why put this in the loop if it's gonna return?

Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
```diff
@@ -434,6 +442,45 @@ private void OnShuttleCallAttempt(ref CommunicationConsoleCallShuttleAttemptEven
...
+                continue;
+
+            MapId? targetStationMap = null;
+            if (nukeops.TargetStation != null && TryComp(nukeops.TargetStation, out StationDataComponent? data))
+            {
+                var grid = data.Grids.FirstOrNull();
+                targetStationMap = grid != null
+                    ? Transform(grid.Value).MapID
+                    : null;
+            }
```
SnappingOpossum, line 467: Is the station entity not in the map?

SnappingOpossum (changes requested):

SnappingOpossum:
Forgot to put it in the review itself but I also feel like it'd be better to give cargo back limited access to guns than do this.

imatsoup:
> Forgot to put it in the review itself but I also feel like it'd be better to give cargo back limited access to guns than do this.

already doing that a little bit with the laser crates and i still think all the problems that previously existed with armory crates are still there, don't really want to regress back to that.

imatsoup:
especially since the issue we are currently having is with one specific strategy with one specific game mode of several.

Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
imatsoup: yeah i'm not sure, trying to get the mapid from the station for comparison didn't work for some reason

CarpBot (changes requested):

PR #458, opened 2026-07-03
Kill round end music

Atakku:
## General information

Killed round end music
Lobby music still exists ofc

## Why / Balance

It can be really ear jarring to hear both round end music & musician playing at the same time, especially so with us pushing towards post-round bar interaction and musicians playing music in them almost every round. So I killed round end music to solve that.

## Technical details

- [x] Commented out one line

## Test plan

Start round. End round. End of round screen pops up but no music starts playing. Restart round now or respawn to lobby. Music plays

## Media

No

## Breaking changes

No

## Changelog

:cl:
- remove: Killed end of round music so musicians can play their music without distractions.


Atakku:
john selfmerg

PR #460, opened 2026-07-04
PDAs lock their uplinks on drop or being stripped

imatsoup:
## General information

Open uplinks will now automatically lock if they are stripped or otherwise dropped by their current holder.
## Why / Balance

Kills metagaming and complaints around PDA checking. Now you only need to do a PDA search for programs or funny ID cards. Traitors can still willingly give their uplink code to someone if they really want (and they can use it on their own PDA if they choose, given that uplinks are no longer tied to specific PDAs).
## Test plan

Make character a traitor, open uplink, drop/strip/throw PDA, the uplink is now closed.
## Changelog

:cl:
- tweak: Open Traitor Uplinks now lock when their PDA is stripped or dropped.

### DIFF
--- a/Content.Server/PDA/PdaSystem.cs
+++ b/Content.Server/PDA/PdaSystem.cs
@@ -65,7 +65,7 @@ namespace Content.Server.PDA
             SubscribeLocalEvent<PdaComponent, InventoryRelayedEvent<ChameleonControllerOutfitSelectedEvent>>(OnRelayedEventToIdCard);
             SubscribeLocalEvent<PdaComponent, InventoryRelayedEvent<VoiceMaskNameUpdatedEvent>>(OnRelayedEventToIdCard);
 
-            SubscribeLocalEvent<PdaComponent, DroppedEvent>(OnDrop); // Carpmosia-edit - Uplink Auto-locker
+            SubscribeLocalEvent<RingerUplinkComponent, DroppedEvent>(OnDrop); // Carpmosia-edit - Uplink Auto-locker
         }
 
         private void OnRelayedEventToIdCard<T>(Entity<PdaComponent> ent, ref InventoryRelayedEvent<T> args)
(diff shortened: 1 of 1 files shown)


Content.Server/PDA/PdaSystem.cs
```diff
@@ -305,6 +308,17 @@ private void OnUiMessage(EntityUid uid, PdaComponent pda, PdaLockUplinkMessage m
             }
         }
 
+        // Carpmosia-start - Uplink Auto-locker
+        private void OnDrop(Entity<PdaComponent> ent, ref DroppedEvent args)
+        {
+            if (!TryComp<RingerUplinkComponent>(ent, out var uplink))
+                return;
+
+            if (uplink!= null && uplink.Unlocked)
```
SnappingOpossum: First part is always true. Really this should be doing the same as the regular uplink locking above, and I'd consider subscribing RingerUplinkComponent to this instead since that should be on all relevant PDAs and only those.

SnappingOpossum (changes requested):
I still don't think this is a good thing to add but I'm not here to be a designer so eh.

Content.Server/PDA/PdaSystem.cs
```diff
@@ -305,6 +308,14 @@ private void OnUiMessage(EntityUid uid, PdaComponent pda, PdaLockUplinkMessage m
             }
         }
 
+        // Carpmosia-start - Uplink Auto-locker
+        private void OnDrop(Entity<RingerUplinkComponent> ent, ref DroppedEvent args)
+        {
+            _ringer.LockUplink(ent.Owner);
```
SnappingOpossum: Reset the store, pass in the component.
```suggestion
            if (TryComp<RemoteStoreComponent>(ent, out var remoteStore))
                remoteStore.Store = null;
            _ringer.LockUplink(ent!);
```

SnappingOpossum (changes requested):

SnappingOpossum (approved):
Does what it's meant to.

Atakku (approved):
mrrp miau

PR #461, opened 2026-07-04
Stranded evac pods

Atakku:
## General information

Instead of FTLing evac pods to CC, makes them end up in some weird places

## Why / Balance

Objectives for escaping on evac pods don't count anyway, and it doesn't make any logical sense for them to make it to CC (they only have thrusters in one direction)
Also players should be motivated to use regular shuttle over pods, so antags can intercept them last minute if absolutely necessary.
Pods should be used as absolute last measure, or by antags who need to survive but can't go by evac shuttle for various reasons (avoiding security)
Pods land on planets like Arrivals and maybe others if possible.
Being stranded in space for the duration of round end is boring so I have decided to NOT implement that.

## Technical details

- [x] Staggered pod launching further for more fun scenarios
- [x] Made pods land on arrivals

## Test plan

Start game on a map with evac pods, get into a pod, callshuttle 0, launchemergencyshuttle, and enjoy the show

## Media

Video is refusing to upload
image

## Breaking changes

No

## Changelog

:cl:
- tweak: Evac pods now have a much larger randomized FTL delay
- tweak: Evac pods leave the FTL space at a randomized delay, before the Evac shuttle arrives to the Terminal
- tweak: Evac pods no longer FTL to the Terminal, instead they land at a random spot on the Arrivals planet.
- tweak: Arrivals cloud radius has been increased to account for the Evac pods.


Raylative:
this idea is epic

SnappingOpossum:
Have you tried to see what happens when they overlap?

Atakku:
> Have you tried to see what happens when they overlap?

no i did not care cuz its quite unlikely and upstream probably uses similar logic to spawning in the shuttles (which occasionally, pretty rarely, end up stuck in stuff)

SnappingOpossum (approved):
Atakku is threatening me.

DahBrass:
john self merge this NOW

PR #479, opened 2026-07-10
Try to improve BuyerAccessCondition

SnappingOpossum:
## General information

Title, mostly just trying to make this more consistent with other conditions.

## Why / Balance

Had some parts that were unused or could be safely replaced with more modern things, felt like I'd try to brush up the rest of it as well.
Partly got rid of the TryGetComponents in case their systems switch to using queries internally.

## Technical details

Tried to store AccessReaderSystem (and now EmagSystem) instead of getting them every time, tried to clarify some things with comments, and added support for things without AccessReaderComponent as long as this has an access set. Also made this default to true when it finds nothing to check against, but log an error when doing so.
Switched from using a TryGetComponent<AccessReaderComponent> to GetMainAccessReader.
Would have switched from TryGetComponent<EmaggedComponent> to CheckFlag but AccessReaderComponent prevents EmaggedComponent from getting added.

## Test plan

Use the VendingStoreSalvage I guess, mess with the Access field on a listing to make sure that works as well.

## Changelog

Shouldn't be noticeable.

### DIFF
--- a/Content.Server/_Carpmosia/Store/Conditions/BuyerAccessCondition.cs
+++ b/Content.Server/_Carpmosia/Store/Conditions/BuyerAccessCondition.cs
@@ -1,51 +1,72 @@
-using System.Linq;
-using Content.Shared.Store;
+using Content.Shared.Access;
 using Content.Shared.Access.Components;
 using Content.Shared.Access.Systems;
-using Content.Shared.Emag.Components;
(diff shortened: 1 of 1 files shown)


imatsoup (approved):
seems to work fine

CarpBot (changes requested):

PR #480, opened 2026-07-10
Regenerated wall sprites

Atakku:
## General information

Part 2 of #440, because I got tired of rewriting the shitcode and took a break.
Reimplemented wallsprite procgen

## Why / Balance

tech debt or smth

## Test plan

No

## Media

No, everything is still the same

## Breaking changes

No

## Changelog

No


Atakku:
i am john selfmerge

PR #482, opened 2026-07-11
Let ghosts return to lobby at Evac & disable late join

Atakku:
## General information

Made it so as soon as evac arrives, or round ends (in case it ended early), ghosts will be able to return to lobby, and latejoin will be disabled to prevent abuse & joining too late

## Why / Balance

We intend to increase the end of round timer from 5 to 7 minutes, and shrink the lobby timer from 5 to 3 minutes to maintain the 10 minute window between rounds while offering more fun for those who don't need to take a break.
Having a respawn button right before the end of round would let ghosts (or those who cryo'd at EoR or station) go back to lobby to adjust their character early (because the new 3 minute timer is very short). And they will have more than 10 minutes as it's unlocked just a little earlier.

## Technical details

- [x] Added a button to return to lobby to the ghost menu
- [x] Made the return to lobby button unlock only once the evac shuttle has arrived.
- [x] Made the game lock out "Late join" button when the evac shuttle arrives

## Test plan

Start round, `callshuttle 0`, and see that the late-join button is disabled, observe to see that the back to lobby button is enabled

## Breaking changes

None

## Changelog

:cl:
- add: Ghosts now have a button to return to lobby early, which unlocks as soon as the evac shuttle arrives.
- tweak: Late-joining is disabled when the evac shuttle arrives.


imatsoup (approved):
conceptually fine.

Content.Client/UserInterface/Systems/Ghost/GhostUIController.cs
```diff
@@ -16,6 +16,7 @@ public sealed partial class GhostUIController : UIController, IOnSystemChanged<G
     [UISystemDependency] private readonly GhostSystem? _system = default;
 
     private GhostGui? Gui => UIManager.GetActiveUIWidgetOrNull<GhostGui>();
+    private bool _canReturnToLobby = false; // Carpmosia-edit - Return to lobby
```
SnappingOpossum: False is the default value.

Content.Server/Ghost/GhostSystem.cs
```diff
@@ -272,6 +273,22 @@ private void OnGhostReturnToBodyRequest(GhostReturnToBodyRequest msg, EntitySess
             _mind.UnVisit(actor.PlayerSession);
         }
 
+        // Carpmosia-start - Return to lobby
+        private void OnGhostReturnToLobbyRequest(GhostReturnToLobbyRequest msg, EntitySessionEventArgs args)
+        {
+            // Checking for post round in case of an early "restartround"
+            if (!_configurationManager.GetCVar(CCVars.GameDisallowLateJoins)
```
SnappingOpossum, line 280: I'm not a fan of tying this to when latejoins are allowed but I don't really care that much if this works as is.

SnappingOpossum (commented):
I have some minor things to say but most of this is out of my depth so I'd not gonna block anything here.

Content.Client/UserInterface/Systems/Ghost/GhostUIController.cs
Atakku: uh do i not need to declare false in c#? or replace it with default!?

Content.Server/Ghost/GhostSystem.cs
Atakku: i need latejoins off at the same time anyway, so its better to have the system work off of one shared signal than have potential issues down the line

Content.Client/UserInterface/Systems/Ghost/GhostUIController.cs
SnappingOpossum: You can just do `private bool _canReturnToLobby;` and it'll be assumed to be false. The `= default!` is specific to non nullable types with a null default value.

Atakku: oh didnt know

PR #490, opened 2026-07-12
Make anomaly generator emag-able

Atakku:
## General information

Made anomaly generator emag-able, which reduces the 5 minute cooldown to 8 seconds (length of animation)

## Why / Balance

Could help with the critting anoms objective, as well as a potential to cause some chaos.

## Technical details

- [x] Made anomaly generator replace CooldownLength with GenerationLength when emagged, also set CooldownEndTime to zero so there isnt a leftover cooldown if generator was used previously.

## Test plan

Start server, go to science, emag the generator and have fun

## Media

Nah

## Breaking changes

No

## Changelog

:cl:
- add: Anomaly generator can now be EMAG-ed to remove the 5 minute safety cooldown.

### DIFF
--- a/Content.Client/Anomaly/Ui/AnomalyGeneratorWindow.xaml.cs
+++ b/Content.Client/Anomaly/Ui/AnomalyGeneratorWindow.xaml.cs
@@ -53,6 +53,12 @@ public sealed partial class AnomalyGeneratorWindow : FancyWindow
     {
         if (_timing.CurTime > _cooldownEnd)
         {
+            // Carpmosia-start - Emagging anomaly generators
+            if (GenerateButton.Disabled)
+            {
+                UpdateReady();
+            }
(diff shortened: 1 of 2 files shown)


imatsoup:
I'm not sure this is the best way to go about it. You could be creating a situation where science is incentivized to work with traitors or thieves to generate anomalies quickly to speedrun points. Maybe a better solution to helping people with the objective would be to have there be a kind of 'anomaly hunting' kit in the uplink under the objectives tab that appears when you have the crit objective. It could have a modified CHIMP, and an anomaly locator (maybe in evil colors?).

SnappingOpossum:
Honestly I think it'd be better to replace the objective or remove it outright.

Atakku:
sure then i will close this, i just kinda quickly made this cuz i had an idea in the todo list

PR #494, opened 2026-07-12
Automatic map vote

Atakku:
## General information

Make map voting automatic, kill manual map votes (for non admins)

## Why / Balance

People calling map votes in middle of the round can be quite annoying as for some people who run on lower resolution it can take up more than half the screen for a very long time.
Also sometimes we forget to run map vote at the start of the round lmao.
Considering we are planning to reduce lobby time (and increase end of round time), this is needed as it can be otherwise too late for most map votes.

## Technical details

- [x] Killed manual map votes.
- [x] Made map voting automatic at start of each round.

## Test plan

Start server
Join server
Map vote should be already running

## Media

No

## Breaking changes

No

## Changelog

:cl:
- tweak: Map vote has been changed to be exclusively automatic at the start of each round, unless started by an Admin.


imatsoup (approved):
seems like a great change

Content.Server/GameTicking/GameTicker.RoundFlow.cs
```diff
@@ -684,6 +688,22 @@ public void RestartRound()
                 UpdateInfoText();
 
                 ReqWindowAttentionAll();
+
+                // Carpmosia-start - Automatic map vote
+                if (_cfg.GetCVar(CCVars.GameLobbyAutoVote))
+                {
+                    // There isn't really a better way to identify an already running map vote...
+                    if (!_voteManager.ActiveVotes.Any(x => x.Title == Loc.GetString("ui-vote-map-title")))
```
SnappingOpossum: Use an All and negate the internal condition instead of negating an Any.
```suggestion
                    if (_voteManager.ActiveVotes.All(x => x.Title != Loc.GetString("ui-vote-map-title")))
```

Content.Server/GameTicking/GameTicker.RoundFlow.cs
```diff
@@ -24,6 +26,7 @@
 using Robust.Shared.Network;
 using Robust.Shared.Player;
 using Robust.Shared.Random;
+using Robust.Shared.Timing;
```
SnappingOpossum: Comment.

SnappingOpossum (changes requested):

Content.Server/GameTicking/GameTicker.RoundFlow.cs
Atakku: oh ye i forgor when i aded the delay

SnappingOpossum (approved):

PR #495, opened 2026-07-12
Extend End-of-Round time & Shorten lobby time

Atakku:
## General information

Requires #482 and #494.
        
Extended end-of-round timer by 150 seconds.
Shortened lobby timer by 150 seconds.
Preserved the between-round break of 10 minutes.

## Why / Balance

This extends the late of round gameplay without affecting the length of the overall 10 minute round break.
#482 will let any ghost (there is cryosleep at CC) to respawn back to lobby at end-of round, therefore we don't need an overly long 5 minute lobby timer anymore, as if someone needs to customize their character or do something between rounds, they can do it by leaving the round early.
Also #494 ensures a map vote happens in time, and isn't started late

## Technical details

Two lines of config changed.

## Test plan

Start server and end round.

## Media

No

## Breaking changes

No

## Changelog

:cl:
- tweak: Lobby time has been shortened from 300 seconds to 150 seconds
- tweak: End of round has been extended from 300 seconds to 450 seconds

### DIFF
--- a/Resources/ConfigPresets/_Carpmosia/production.toml
+++ b/Resources/ConfigPresets/_Carpmosia/production.toml
@@ -13,8 +13,8 @@ rules.time = 90
 discord.news_webhook_send_during_round = true
 
 # CCVars.Game
-game.lobbyduration = 300
-game.round_restart_time = 300
+game.lobbyduration = 150
+game.round_restart_time = 450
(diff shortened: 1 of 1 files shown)


imatsoup (approved):
makes very little difference to me

PR #517, opened 2026-07-19
Little tiny "Dale" Jr. basic mob

Jakeybobble:
## General information

"Dale" Jr. is the son of Dale. They are non-aggressive unlike their parent and have a little happiness inside of themselves.  
This PR contains only the basic mob so it is available for future gameplay features, but only exists for admeme otherwise.

## Why / Balance

Dale Jr. is just a silly little guy!  
Could be fun admeme on exo.  

## Technical details

- Created `MobXenoDaleJunior` prototype.
- Added `young_praetorian.rsi` sprites.

## Test plan

Make sure changes from parent apply.  
NOTE: For some reason, without `offset: 0,0` in the sprite component, weird stuff happens.

## Media

Screenshot_20260720_001355
dale_jr_spin

## Changelog

Nothing right now, since Dale Jr. can't be spawned without admeme.


Atakku (changes requested):
tbh feels a little too gray, shouldn't it be shaded darker to look more like dale?

Jakeybobble:
> tbh feels a little too gray, shouldn't it be shaded darker to look more like dale?

I'll doodle around a bit more, see if I can make it darker/glossier.

CarpBot (changes requested):
Awaiting approval from @GlupiyBrush, @Atakku

PR #526, opened 2026-07-25
Adjusted some timers

Atakku:
## General information

Increased lobby timer back to 300 seconds
Also doubled map vote duration because why not.

## Why / Balance

People like the longer EoR change, but they also would like more lobby time as well.

## Technical details

2 config changes

## Test plan

Start server and join lobby

## Media

Long vote
image

## Breaking changes

No

## Changelog

:cl:
- tweak: Lobby duration has been changed back from 150 seconds to 300.
- tweak: Map vote duration has been doubled from 90 seconds to 180.


CarpBot (approved):

PR #538, opened 2026-07-28
Restart vote tweaks

imatsoup:
## General information

Lower the restart round threshold for restart votes to 65% as a test. Remove restart votes automatically failing if an admin is online (because why?)
## Why / Balance

Restart round threshold is too high and the condition of failing on an admin being present is just silly. The new threshold may be too low but we can incrementally increase it if it is abused.
## Changelog
:cl: 

- tweak: The threshold of votes required for a round restart vote to pass has been reduced.
- tweak: Round restart votes will no longer automatically fail if there is an admin present.

### DIFF
--- a/Resources/ConfigPresets/_Carpmosia/common.toml
+++ b/Resources/ConfigPresets/_Carpmosia/common.toml
@@ -45,3 +45,5 @@ shuttle.arrivals_returns = true
 vote.preset_enabled = false
 vote.map_enabled = false
 vote.timermap = 180
+vote.restart_required_ratio = 0.65
+vote.restart_not_allowed_when_admin_online = false;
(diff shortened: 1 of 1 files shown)


CarpBot (changes requested):

Resources/ConfigPresets/_Carpmosia/common.toml
```diff
@@ -45,3 +45,5 @@ shuttle.arrivals_returns = true
 vote.preset_enabled = false
 vote.map_enabled = false
 vote.timermap = 180
+vote.restart_required_ratio = 0.65
+vote.restart_not_allowed_when_admin_online = false;
```
Atakku:
```suggestion
vote.restart_not_allowed_when_admin_online = false
```
this aint how this works!!!!

Atakku (changes requested):

Atakku (approved):
mrrp miau

CarpBot (approved):

PR #541, opened 2026-07-29
Rebalanced job weights

Atakku:
## General information

Changed job spawn weights from:
- Captain and HoP
- Rest of command
- Rest of jobs

To something more organized like:
- Captain
- HoP
- HoS
- StationAI
- QM, CMO, CE, RD
- Warden
- Atmospheric Technician, Chemist, Detective
- Paramedic
- Security Officer, Medical Doctor
- Security Cadet, Medical Intern
- Salvage Specialist
- Cargo Technician, Station Engineer, Scientist, Borg
- Technical assistant, Research assistant
- Botanist, Chef
- Bartender, Janitor, Lawyer, Psychologist, Reporter
- Service worker
- Chaplain, Librarian, Musician
- Clown, Mime
- Passenger

## Technical details

- [x] Adjusted weights of pretty much all jobs

## Test plan

Start game and check job order

## Media

No

## Breaking changes

No

## Changelog
:cl: 

- tweak: Rebalanced job roll order.

PR #542, opened 2026-07-29
Killed job weights with hammers

Atakku:
## General information

An opposite PR of #541, instead of making intricate hierarchy, just killed all job weights to 0

## Why / Balance

Maybe to stop the constant tug of war between the game and the players, we should just cut the rope

I've heard some feedback that currently as is, people just **avoid** setting command roles on **low** at all because they pretty much **never** roll their other jobs.

This will take time for people to adapt to, because a lot of people just have command unselected for the very reason that it steals priority from everything else. But this should in longterm resolve the confusion, and make job rolls much more straight forward.

## Technical details

- [x] Commented out all weight overrides
- [x] Adjusted related tests

## Changelog
:cl: 

- remove: Removed command role prioritization when rolling player jobs

### DIFF
--- a/Content.IntegrationTests/Tests/Round/JobTest.cs
+++ b/Content.IntegrationTests/Tests/Round/JobTest.cs
@@ -157,7 +157,7 @@ public sealed class JobTest : GameTest
         await pair.Server.WaitPost(() => ticker.StartRound());
         await pair.RunTicksSync(10);
 
-        AssertJob(pair, Captain);
+        AssertJob(pair, Engineer); // Carpmosia-edit - Job priority rebalance
 
         await pair.Server.WaitPost(() => ticker.RestartRound());
     }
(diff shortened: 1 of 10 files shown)


CarpBot (changes requested):
Pending triage

Content.IntegrationTests/Tests/Round/JobTest.cs
```diff
@@ -149,15 +149,15 @@ public async Task JobWeightTest()
...
+        Assert.That(captain.Weight, Is.EqualTo(engineer.Weight)); // Carpmosia-edit - Job priority rebalance
         Assert.That(engineer.Weight, Is.EqualTo(passenger.Weight));
 
         await pair.SetJobPriorities((Passenger, JobPriority.Medium), (Engineer, JobPriority.High), (Captain, JobPriority.Low));
         ticker.ToggleReadyAll(true);
         await pair.Server.WaitPost(() => ticker.StartRound());
         await pair.RunTicksSync(10);
 
-        AssertJob(pair, Captain);
+        AssertJob(pair, Engineer); // Carpmosia-edit - Job priority rebalance
```
SnappingOpossum: You're just making this another check of job priority add a test proto or make the test explicit.

CarpBot (approved):

PR #543, opened 2026-07-29
More job priorities

Atakku:
## General information

Added more job priorities because Low and Medium kinda fucking sucks as its just a binary choice between maybe and not really

## Why / Balance

In tandem with #542, we can let players choose the jobs themselves, self organize and potentially enable all the jobs roundstart. The more players ready up with more jobs selected, the more likely the game just balance out all the jobs according to population (map prototype), and have all the roles satisfied, without people feeling like "rolling X 5th time in a row" unless they want to ofc, keeping gameplay more varied as people join/leave with different job priorioties

## Technical details

- [x] Added 3 more priority selections
- [x] Halved the size of the priority buttons to fit everything nicely (4 -> 6)
- [x] Renamed existing priority selections to fit with the rest nicer

## Test plan

Start game, go pick jobs for character

## Media

FROM: (ignore the order)
image

TO:
image

## Breaking changes

No

## Changelog
:cl: 

- tweak: Job priorities have been subdivided into 6 total selections.

### DIFF
--- a/Content.Server.Database/Model.cs
+++ b/Content.Server.Database/Model.cs
@@ -371,13 +371,11 @@ namespace Content.Server.Database
         // These enum values HAVE to match the ones in JobPriority in Content.Shared
         Never = 0,
         // Carpmosia-start - More job priorities
-        Low5 = 1,
-        Low4 = 2,
-        Low3 = 3,
-        Low2 = 4,
-        Low = 5,
-        Medium = 6,
-        High = 7
+        Lowest = 1,
+        Lower = 2,
+        Low = 3,
+        Medium = 4,
+        High = 5
         // Carpmosia-end - More job priorities
     }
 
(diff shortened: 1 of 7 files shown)


CarpBot (changes requested):
Awaiting approval from: @SnappingOpossum

Content.Server.Database/Model.cs
```diff
@@ -370,9 +370,13 @@ public enum DbJobPriority
...
         // These enum values HAVE to match the ones in JobPriority in Content.Shared
         Never = 0,
-        Low = 1,
-        Medium = 2,
-        High = 3
+        // Carpmosia-start - More job priorities
+        Lowest = 1,
+        Lower = 2,
+        Low = 3,
+        Medium = 4,
```
SnappingOpossum, line 377: Something like this should work fine no?
```suggestion
        Min = 1,
        Low = 2,
        Medium = 3,
        MidHigh = 4,
```

Atakku: high would reset to medium, from 5 to 3 which is the issue

SnappingOpossum (approved):
Just turn off my review being required it genuinely doesn't matter whether this has my approval or not.

CarpBot (approved):

PR #555, opened 2026-08-03
Spin particles

P3tux0:
## General information
Made Omega particles slightly more saturated and visually changed the lighting from all anomalous particles to make it look like they are spinning

## Why / Balance
Omega particles is halfway in the spectrum of Epsilon particles and i think minor color blindness would have issue distinguishing the two, also Omega dealing 4 times the heat damage from regular particles means it should be brighter

As to why make them spin: It just looks cool

## Technical details
Added a new sprite to Textures/_Carpmosia/Effects/LightMasks

## Test plan
Get a Syndicate C.H.I.M.P and check out the particles in a dark place

## Media
image
image
image
image
image

## Breaking changes
none, purely visual

## Changelog
:cl:
- tweak: Made Omega particles slightly more green to remove confusion with Epsilon particles
- tweak: Changed the lighting from Anomalous particles to make it look like they are spinning


CarpBot (changes requested):
Awaiting approval from: @Atakku, @GlupiyBrush

PR #569, opened 2026-08-06
Configurable BleatingAccent and sheep/goat accent change

ultrafiche:
## General information

BleatingAccentComponent gives an entity an accent that lengthens certain consonant vowel pairs, this PR makes that length configurable and allows for an optional "stop" to be inserted just before the end of the extension as a kind of glottal stop sort of thing goats and sheep make weird noises ok

## Why / Balance

I think the defaults I've set it to (3 vowels with a apostrophe, "baa'ah") are more in line with how I imagine a goat would talk. Also the future ability to give Ferrodae a slightly weakened version of this accent as having to taaaalk liiiike thiiiis as a player would be annoying.

## Technical details
BleatRegex is now generated from an attribute for the love of the game. BleatingAccentSystem has a bit more string building stuff in it and BleatingAccentComponent has two more fields for configuration.
## Test plan

spawned in and cogni'd both a goat and sheep on a localhost, said some goat-esque nonsense

## Media

https://github.com/user-attachments/assets/fd0f064c-a534-4207-9c9a-fbe736c60c3a

## Breaking changes
Nothing breaking.

## Changelog
:cl: 

- tweak: Cognizined goats and sheep have a slightly different accent.

### DIFF
--- a/Content.Server/Speech/EntitySystems/BleatingAccentSystem.cs
+++ b/Content.Server/Speech/EntitySystems/BleatingAccentSystem.cs
@@ -1,17 +1,27 @@
+using System.Linq;
 using System.Text.RegularExpressions;
 using Content.Server.Speech.Components;
 using Content.Shared.Speech.EntitySystems;
+using Robust.Shared.Random;
 
 namespace Content.Server.Speech.EntitySystems;
 
 public sealed partial class BleatingAccentSystem : RelayAccentSystem<BleatingAccentComponent>
 {
-    private static readonly Regex BleatRegex = new("([mbdlpwhrkcnytfo])([aiu])", RegexOptions.IgnoreCase);
+    [Dependency] private IRobustRandom _random = default!;
+
+    [GeneratedRegex("([mbdlpwhrkcnytfo])([aiu])", RegexOptions.IgnoreCase, "en-US")]
+    private static partial Regex BleatRegex();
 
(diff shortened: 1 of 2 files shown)


CarpBot (changes requested):

Content.Server/Speech/EntitySystems/BleatingAccentSystem.cs
```diff
@@ -6,12 +7,20 @@ namespace Content.Server.Speech.EntitySystems;
...
     {
-        // Repeats the vowel in certain consonant-vowel pairs
-        // So you taaaalk liiiike thiiiis
-        return BleatRegex.Replace(message, "$1$2$2$2$2");
+        // Carpmosia-start - configurable bleating
+        // Repeats the vowel in certain consonant-vowel pairs and adds a stop near the end.
+        // So you taaa-alk liii-ike thiii-is
+        var repeats = ent.HasValue ? ent.Value.Comp.Repeats : 3;
+        var stop = ent.HasValue ? ent.Value.Comp.Stop : "'";
+        var replacement = "$1" + string.Concat(Enumerable.Repeat("$2", repeats - 1)) + stop + "$2";
```
ultrafiche, line 22: With what I know this was the most elegant way of doing the vowel repeating that I could come up with. If there's a nicer way possibly without linq I'm open to it.

PR #577, opened 2026-08-08
Update evac launch announcement

imatsoup:
## General information

Change the announcement made when the evac shuttle is launched to say we are going to the Terminal and not to Centcomm.
## Why / Balance

Its different so it should be different.
## Technical details

Changed the local string to point to our own thing.

## Test plan

Load up map. Callshuttle 1. When it arrives, launch the shuttle from the console. The message is different.
## Changelog
:cl: 

- tweak: Tweaked the emergency shuttle launch announcement to no longer erroneously state the shuttle is going to Centcomm.

### DIFF
--- a/Resources/Locale/en-US/_Carpmosia/shuttles/emergency.ftl
+++ b/Resources/Locale/en-US/_Carpmosia/shuttles/emergency.ftl
@@ -1,0 +1,1 @@
+emergency-shuttle-left-carp = The Emergency Shuttle has left the station. Estimate {$transitTime} seconds until the shuttle arrives at the Terminal.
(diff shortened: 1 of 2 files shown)


CarpBot (changes requested):

Resources/Locale/en-US/_Carpmosia/shuttles/emergency.ftl
```diff
@@ -0,0 +1 @@
+emergency-shuttle-left-carp = The Emergency Shuttle has left the station. Estimate {$transitTime} seconds until the shuttle arrives at the Terminal.
```
Atakku, line 1: maybe at a terminal is better cuz there is no specific Terminal
```suggestion
emergency-shuttle-left-carp = The Emergency Shuttle has left the station. Estimate {$transitTime} seconds until the shuttle arrives at a Terminal.
```

Atakku (changes requested):

PR #583, opened 2026-08-12
Better map rotation

Atakku:
## General information

Improved map rotation by making map rotation & map voting cross-compatible
CurrentlyEligibleMaps now returns a list sorted based on weight (from map rotation), moving already played maps to the end of the list
Map votes now consider only top N maps available (so during lowpop maps are still available, but during high pop the previous choices are filtered out)

## Why / Balance

Force people to actually cycle maps instead of 500 amber rounds, while still preserving map votes

## Technical details

- [x] Patched map rotation/map vote to work together

## Test plan

Comment out `map.MinPlayers <= _playerManager.PlayerCount &&` in `Content.Server/Maps/GameMapManager.cs:213`
Switch map pool to `UpstreamMapPool` or smth (Carpmosia map pool is cooked rn)
Wait for map vote to start
Let map vote pass
startround
restartroundnow
Repeat from beginning and see map not present

## Media

First round (i chose snowball)
image
Second round
image
Third round (i chose plasma)
image
Fourth round
image

## Changelog
:cl: 

- tweak: Map vote options count has been reduced to 3
- tweak: Map vote will now try to hide maps previously voted

### DIFF
--- a/Content.Server/Maps/GameMapManager.cs
+++ b/Content.Server/Maps/GameMapManager.cs
@@ -98,7 +98,11 @@ public sealed partial class GameMapManager : IGameMapManager
 
     public IEnumerable<GameMapPrototype> CurrentlyEligibleMaps()
     {
-        var maps = AllVotableMaps().Where(IsMapEligible).ToArray();
+        // Carpmosia-start - Better map rotation
+        var maps = AllVotableMaps().Where(IsMapEligible)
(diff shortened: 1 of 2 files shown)


CarpBot (changes requested):
Pending triage

CarpBot (approved):

PR #585, opened 2026-08-13
Temporarily replace Domain with Sacrifice

imatsoup:
## General information

Replaces Rat King's Domain ability with a new ability that allows them to sacrifice rat servants for a heal.
## Why / Balance

Domain is kind of a trash ability for its normal use-case. Since the heal is based on the concentration of ammonia, most rooms are too open to get much healing from it. It *does* heal rat servants too, but most of the time the line between an injured rat servant and a dead rat servant is very thin. Where the ability DOES excel is when used underneath an airlock, where the entire ammonia cloud concentrates and just about fully heals the Rat King. This feels like a borderline exploit to me, I don't really have the bandwith to think about how to change Domain to 'fix' this, and in the end we would just be taking a bad ability with a niche use case and remove its one use case.

Sacrifice is a much simpler ability: You trade one rat servant's life for an instant, distributed heal (on a decent cooldown). Its not particularly exciting, but at least it puts the focus back on the rat king building a horde of rats rather than the weird lone-rat strategy employed by people using Domain to it's full effect.
## Technical details

- [x] Implemented Z and X to handle Y edge case.
- [X] Added a new component for all X to prevent Z.

## Test plan

Load up game, spawn rat king, hurt the rat king, take over the rat king, spawn a rat and then use sacrifice, observe your healed wounds.
## Media

https://github.com/user-attachments/assets/8402c6e6-7e32-4758-96f7-2fc92b1a3863

## Breaking changes

probably not
## Changelog
:cl: 

- remove: Temporarily disabled Rat King's Domain ability due to it's interactions with airlocks
- add: Added Sacrifice ability to Rat King, allowing them to trade rat servants for a heal


CarpBot (changes requested):

imatsoup:
our very own artmaint has provided us with a nice little icon for the ability, which should help communicate to people its been changed 
ratKingSacrifice

GlupiyBrush (approved):
beautiful sprite

PR #608, opened 2026-08-27
Cash is currency

CDWimmer:
## General information

behind the scenes changes to let cash be used as currency for future vending stores.

## Technical details

Add Currency component to SpaceCash (is this correct usage?)
Add new instance of currency type for SpaceCash (Should this id be named something else? Credit?)
Add ftl for proper display in vending stores, should Speso(s) be the currency used. 

## Changelog
:cl:


CarpBot (changes requested):

PR #609, opened 2026-08-27
Wallets

CDWimmer:
## General information

Adds wallet(s). Let's you keep money, salvage tickets, TC, and bits of paper if you really want. 
(Strictly speaking stores any small item with the "Currency" Component, or Paper component)

They can be printed from a uniform printer, emag required for Syndicate wallet. 

## Why / Balance

It would be fun! 

## Technical details

- [x] Add wallet entity, a small storage item that fits in pockets and stores currency items
  - For Speso, requires previous PR adding Currency to SpaceCash. 
- [x] Derivative forms wallet, such as purse, and possibly themed after other factions.
- [x] Sprites (icons + handheld)
- [x] sprite layers depending on content
- [x] Uniform printer recipes
- [x] Uniform printer wallet category

## Test plan

Spawn wallet,
try putting it in belt slot and pockets, 
try putting it in a backpack to see the 2x2 shape, despite being a small item (defaults to 1x2),
place a stack of salvage tickets inside,
place salvage tickets inside,
place paper-derived items inside (Paper, notes, etc.) inside,
place telecrystals inside.
Provide a uniform printer appropriate materials, print each wallet type
Spawn a few of the pre-filled prototypes and check the randomised contents looks as expected

## Media

image
image

## Changelog
:cl: 

- add: Wallets! To store money, tickets, paper, etc.

PR #622, opened 2026-09-16
Wallet Slot

CDWimmer:
## General information

Depends on previous PR: #609.
Adds an inventory slot for wallets. 

## Why / Balance

~~Because I'm a hack-fraud and stole the idea from Frontier~~ Because if these are going to be things that players keep on their person a lot, they shouldn't have to sacrifice the belt or a pocket slot for it. 

## Technical details

- [x] Add `WALLET` into SlotFlags.cs
- [x] Add new textures to each of the themes for the wallet slot
- [x] Modify playable character species' inventory templates to add the slot to the the clothing slots UI area and stripping UI. 
- [x] As above for Monkeys (and so Kobolds, Kittis), and Scurrets. 

## Test plan

Boot game, control regular playable species, scurret, or monkey/kobold/kitti, and observe the existence of the new slot with the new texture(s). 
Try to place a wallet into the slot without wearing a jumpsuit -> observe it doesn't allow it (slot requires jumpsuit, same as ID slot and pockets)
Try to place wallet into the slot whilst wearing a jumpsuit -> observe it places. 

## Media

image
image

## Breaking changes

none probably?

## Changelog
:cl: 

- add: Inventory slot for wallets

### DIFF
--- a/Resources/Locale/en-US/_Carpmosia/store/currency.ftl
+++ b/Resources/Locale/en-US/_Carpmosia/store/currency.ftl
@@ -2,3 +2,7 @@ store-currency-display-salvage-ticket = {$amount ->
     [1] ticket
     *[other] tickets
 }
+store-currency-display-space-cash = {$amount ->
+    [1] speso
+    *[other] spesos
+}
--- a/Resources/Prototypes/Entities/Objects/Misc/space_cash.yml
+++ b/Resources/Prototypes/Entities/Objects/Misc/space_cash.yml
@@ -5,6 +5,11 @@
   description: You gotta have money.
   components:
   - type: Cash
+  # Carpmosia-start: cash is currency
+  - type: Currency
+    price:
+      SpaceCash: 1
+  # Carpmosia-end: cash is currency
   - type: Item
     shape:
     - 0,0,1,0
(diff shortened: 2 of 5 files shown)


CarpBot (changes requested):
Awaiting approval from: @GlupiyBrush, @Atakku

Content.Shared/Inventory/SlotFlags.cs
```diff
@@ -23,7 +23,9 @@ public enum SlotFlags
     GLOVES = 1 << 10,
     IDCARD = 1 << 11,
     POCKET = 1 << 12,
-    LEGS = 1 << 13,
+    //  Carpmosia-start - legs aren't used, reduce reuse recycle for wallet slot
```
SnappingOpossum: Start comments should always have an end comment to mark the end of changes, use a simpler name for the comment. You've used just `- wallet` for the rest of the pr do that here as well.

SnappingOpossum: Oh also since this isn't a `short` you can just add a new flag without removing one.

CDWimmer: Oh it IS big enough to fit another flag? Idk why I was convinced it a short. How many can we have?

SnappingOpossum: A total of 32, since the default backer is an int. Even if that's reached upping it to a long isn't that big an issue probably

CDWimmer:
holy hell the tests actually passed

